PR #109, opened 2017-09-22
Made the @set eorzeatime A GM Command

AcedArmy:
The @set eorzeatime Command is now //gm time <Value>


takhlaq:
we should change the command prefix to `!` to avoid tagging people on github

PR #115, opened 2017-09-30
Basic invincibility

amibu01:


SapphireMordred:
I am gonna merge this but we need to further investigate what hp update packet is used under which circumstances. If we just send out status update packets with every hp change we might end up with undesired effects.

PR #125, opened 2017-10-09
Player title implementation;

itsMaru:
## Player title implementation ##

Full implementation of the title system for players (adding titles, listing titles, setting titles, SQL sync, player spawn param, sanity check to avoid injection for future use).

### How to use ###

Add your title by using the command "!add title [titleId]". Replace [titleId] with the desired title's ID, then open up the title list in your character's profile and select the title.

javaw_2017-10-09_01-54-21

ffxiv_dx11_2017-10-09_01-55-22

The title add command grabs the title ID value and bitmagic flag it in the titleList property of the Player, while setting the sync flag up for the SQL database.

The packet for the title list request seems to be simply a bitmask (much like QuestCompleteFlags etc).
Its size matches exactly all of the index values in the title exd, without any paddings beyond the header.

The packet for setting the title from the client is just the titleID inside what we use as "param1" of the action handler. We grab that, check if player has the requested title id (pointless for now?), and if valid applies the title as well as set the sync flag up.

### Notes ### 

A weak link I noticed while cooking this up was how if more and more titles get introduced, eventually we're gonna have to change sizes for stuff up in zone/common/rest, alter tables etc.
Perhaps something can be done about it (something minimal like constant variables in Common)? This doesn't affect just tables.

Is there anything we can do to make this better? Thoughts appreciated.


src/servers/Server_Zone/Actor/Player.cpp
```diff
@@ -1420,10 +1420,39 @@ void Core::Entity::Player::setIsLogin( bool bIsLogin )
    m_bIsLogin = bIsLogin;
 }
 
+uint8_t * Core::Entity::Player::getTitleList()
+{
+   return m_titleList;
+}
+
+uint16_t Core::Entity::Player::getTitle()
```
takhlaq: `getTitle() const`
since this function doesnt modify the object

src/servers/Server_Zone/Actor/Player.h
```diff
@@ -328,7 +328,13 @@ class Player : public Actor
    void teleport( uint16_t aetheryteId, uint8_t type = 1 );
    /*! prepares zoning / fades out the screen */
    void prepareZoning( uint16_t targetZone, bool fadeOut, uint8_t fadoutTime = 0, uint16_t animation = 0 );
-   /*! change player's title */
+   /*! get player's title list (available titles) */
+   uint8_t * getTitleList();
+   /*! get player's active title */
+   uint16_t Core::Entity::Player::getTitle();
```
takhlaq: `getTitle() const`
since this function doesnt modify the object

src/servers/Server_Zone/Actor/PlayerSql.cpp
```diff
@@ -316,6 +321,12 @@ void Core::Entity::Player::createUpdateSql()
    if( m_updateFlags & PlayerSyncFlags::HowTo )
       charaDetailSet.insert( " HowTo = UNHEX('" + std::string( Util::binaryToHexString( static_cast< uint8_t* >( m_howTo ), sizeof( m_howTo ) ) ) + "')" );
 
+   if ( m_updateFlags & PlayerSyncFlags::Title )
+   {
+      charaDetailSet.insert( " ActiveTitle = " + std::to_string( m_title ) );
+      charaDetailSet.insert( " TitleList = UNHEX('" + std::string( Util::binaryToHexString( reinterpret_cast< uint8_t* >( m_titleList ), sizeof( m_titleList ) ) ) + "')" );
```
takhlaq: not sure the additional std::string is needed here if Utils::binaryToHexString returns one

src/servers/Server_Zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -257,13 +257,6 @@ void Core::DebugCommandHandler::set( char * data, Core::Entity::PlayerPtr pPlaye
       pPlayer->sendModel();
       pPlayer->sendDebug( "Model updated" );
    }
-   else if ( subCommand == "title" )
-   {
-      uint32_t titleId;
-      sscanf( params.c_str(), "%d", &titleId );
```
takhlaq: %u since unsigned integer

src/servers/Server_Zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -311,6 +304,14 @@ void Core::DebugCommandHandler::add( char * data, Core::Entity::PlayerPtr pPlaye
 
       pPlayer->addStatusEffect( effect );
    }
+   else if ( subCommand == "title" )
+   {
+      uint32_t titleId;
+      sscanf( params.c_str(), "%d", &titleId );
```
takhlaq: %u since unsigned integer

PR #135, opened 2017-10-13
Web template changes for libcef launcher changes

GokuWeedLord:
Not really anything to note other than exposed methods use camel case rather than pascal case now so that was updated alongside a minor change to the logo appearance


GokuWeedLord:
actually I've just realised that I can just change it so it uses the same case as the methods in `WebScriptApi` so this could be closed

¯\_(ツ)_/¯

PR #144, opened 2017-10-19
Added Version and Git Hash to Lobby, Zone Startup Info Banners

AcedArmy:
- Added the Version and Git Hash to the Lobby and Zone Startup Banners
_I know it sounds useless but it should help in the future so we know what version of lobby/zone is being used for support/bug reports_


takhlaq:
it doesnt sound useless at all
it'll a ton when debugging issues

takhlaq:
while you're at it, add a gm command to log current git hash to chat too
`!get revision` or something

PR #145, opened 2017-10-19
Fixing some type alignments via casting

ShelbyZ:
- static_cast for reasonable values
- at least one stray float


SapphireMordred:
While fixing that kinda stuff we should also ensure style.
meaning spaced after < and before > in template types. Other than that, very nice to have someone doing some compiler warning cleanups 👍

PR #153, opened 2017-10-22
 Correctly applying /ZI for debug 

ShelbyZ:
- /ZI is not a link flag
- /ZI is not settable via CMAKE_CXX_FLAGS
- /Zi is default unless add_defitions(/ZI)


ShelbyZ:
Seems there are some weird issues still

PR #157, opened 2017-10-23
More static_cast

ShelbyZ:
- expand some types
- gnarly std::pair issue with short to int8_t
- getLinkshellById now uses uint32_t

### DIFF
--- a/sql/infolinkshell.sql
+++ b/sql/infolinkshell.sql
@@ -23,7 +23,7 @@ DROP TABLE IF EXISTS `infolinkshell`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `infolinkshell` (
-  `LinkshellId` int(20) NOT NULL AUTO_INCREMENT,
+  `LinkshellId` bigint(20) NOT NULL AUTO_INCREMENT,
   `MasterCharacterId` int(20) DEFAULT NULL,
   `CharacterIdList` blob,
   `LinkshellName` varchar(32) DEFAULT NULL,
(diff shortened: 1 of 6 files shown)


SapphireMordred:
LinkshellId was intended to use uint64_t as thats what retail uses aswell.

SapphireMordred:
Int 20 is able to hold int64. The map however sure is wrong in that case.

ShelbyZ:
Unless I'm reading the [documentation](https://dev.mysql.com/doc/refman/5.7/en/integer-types.html) wrong it looks like mysql int(x) only holds 4 bytes with x representing a fill-width.  I also push this clarification because we are pulling mysql response as:

```cpp
uint32_t linkshellId = field[0].get< uint32_t >();
```

and had not had a chance to look at how field retrieval work, if read 64 would step on to the next, etc.

SapphireMordred:
Yes you are actually correct. It needs to be bigint in mysql. I guess that is an oversight in quite some fields.

sql/infolinkshell.sql
```diff
@@ -23,7 +23,7 @@ DROP TABLE IF EXISTS `infolinkshell`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `infolinkshell` (
-  `LinkshellId` int(20) NOT NULL AUTO_INCREMENT,
+  `LinkshellId` bigint(20) NOT NULL AUTO_INCREMENT,
```
takhlaq: ``` `LinkshellId` bigint(20) UNSIGNED NOT NULL AUTO_INCREMENT,```

though at some point we need to do the same for all ids (e.g. MasterCharacterId on the line below)

SapphireMordred: Keep in mind there is currently a sql rewrite ongoing

PR #167, opened 2017-11-06
Add a struct def of job gauge

perize:


GokuWeedLord:
probably worthwhile to build the structs for each class/job difference where required and then throw them in a `union`

PR #169, opened 2017-11-17
Packet definitions; Item level; (WIP)

itsMaru:
## Packet definitions ##

Updating some packets as well as adding Role Actions to PlayerClassInfo.
Note that invalid action IDs crash the client.

## Item level ##

Item level support has been added. Sums up all equipped item's ilvls and then divide it by 12.

Calculated and then sent after a change to equipment is done and an update is due, and on initializing, matching retail. 

Value is clamped range (0,9999) due to possible client crashes if a value bigger than 9999 is sent.

image

Known issue is equipping a newly added item (through //gm item) having weird values read in the calculation from the item info.
Further investigation ongoing.


src/servers/Server_Zone/Inventory/Inventory.cpp
```diff
@@ -439,6 +439,30 @@ bool Core::Inventory::removeCrystal( CrystalType type, uint32_t amount )
...
+      case ItemCategory::CnjWep:
+      case ItemCategory::CrpPri:
+      case ItemCategory::FshPri:
+      case ItemCategory::GlaWep:
+      case ItemCategory::GldPri: // Goldsmith
+      case ItemCategory::LtwPri:
+      case ItemCategory::MinPri:
+      case ItemCategory::ThmWep:
+      case ItemCategory::WvrPri:
+      case ItemCategory::BlmPri: // Blacksmith
```
takhlaq: iirc in game it's BSM for blacksmith

itsMaru: Plenty of the values in the enum are mismatched. I'm waiting on the enum autogeneration however - since this ItemCategory doesn't actually reflect ItemUICategory.

itsMaru:
Hold on to this for a sec - generated enum stuff might come in handy for EquipSlotCategory (currently what's returned for getCategory method on Item class) and UIItemCategory (what we're gonna need to implement for verifying if a weapon is one-handed).

SapphireMordred:
Whats the state on this now?

itsMaru:
Need an elegant way to pull ItemUICategory for an item (stuff like Two-handed Conjurer's Arms etc), as we need that info to check if a weapon is one-handed or not. Does the generator tools help in this case?

I'm also not sure fixing the //gm item bug falls within the scope of this PR

PR #171, opened 2017-11-18
Fix generated enum build break; Remove Mono as dependency from README;

itsMaru:
## Fix generated enum build break ##

Include stdint.h so we can have the type aliases needed (mostly uint8_t)

## Remove Mono as dependency from README ##

None of the tools in this repository depend on a C# runtime.


ayyaruq:
Isn't it `#include <cstdint>` for C++?

itsMaru:
That'd be the standard lib yes, but it seemed to me that we use stdint.h in the project more often.
The idea of consistently using cstdint is definitely interesting and something we should talk about, but would prefer to use namespaces for them (eg std::uint8_t) if doing so.

Thoughts?

ayyaruq:
I'm only decent at C so no clue on namespacing tbh. If it makes code consistent, clearer, and more maintainable though, sounds good.

PR #173, opened 2017-11-19
naive mysql ping timeout (it's dumb, but i guess it works)

takhlaq:


SapphireMordred:
We'll look for a better solution

PR #174, opened 2017-11-21
Reorganize math classes; Fix crash on spawning actor; Style;

itsMaru:
## Reorganize math classes ##

Gave them the Math namespace, put them in their own folder as well as decoupled stat-related calculation from CalcBattle (to CalcStats).

## Fix crash on spawning actor ##

Spawning an actor using the debug command `!add spawn <id> <name_id>` would crash the zone server instantly. This was related to the status effect container not being initialized while the update timer function dereferenced a property.

## Style ##

Reorganized GMCommand handlers so they're in order (also added Kick switch case, although not implemented).


src/servers/Server_Zone/Actor/BattleNpc.cpp
```diff
@@ -66,7 +66,7 @@ Core::Entity::BattleNpc::BattleNpc( uint32_t modelId, uint32_t nameid, const Com
    m_currentStance = Stance::Passive;
 
    m_class = ClassJob::Gladiator;
-   m_level = level > 0 ? level : 60;
+   m_level = level > 0 ? level : 70;
```
takhlaq: maybe in Actor.h define MAX_LEVEL or something

PR #177, opened 2017-11-21
Add a perform handler

perize:
Server just replicates 32bytes of message data from the client.


SapphireMordred:
Does not build -> does not get merged ;)

perize:
@SapphireMordred Should be fixed now.

PR #179, opened 2017-11-22
fix UPDATE_DATE in charainfo not being updated

GokuWeedLord:
absolutely trash tier code


SapphireMordred:
I may have to look into a setTimestamp function for prepared statements. That would eliminate the issue

GokuWeedLord:
should have actually fixed this properly now thanks to dude22072 enlightening me lol

PR #194, opened 2017-12-07
fix character quest data not being removed on character delete

GokuWeedLord:


SapphireMordred:
Db-transaction would be really useful here.

PR #196, opened 2017-12-09
fix build failures

GokuWeedLord:
how does CI even work? witchcraft?


SapphireMordred:
This is not an error, server zone root for has been added as include path, so paths can be relative to that.

GokuWeedLord:
I wonder if cmake cache was messed up and that's why it wouldn't build for me then, works fine now

PR #199, opened 2017-12-11
Native Scripting

GokuWeedLord:
This is more in a code review stage than anything else - more looking for general feedback and whatnot at this point

things to do still:

- [x] make loading _actually work_
- [x] add unloading
- [x] debug commands to load/unload/rebuild scripts
- [ ] port over all the old chaiscript to the new scripts
- [ ] exception/error handling inside scripts
- [ ] proper exporting of used objects/functions instead of exporting everything - nb: everything is exported on linux, is this something we should care to fix or just leave as is?
- [x] generally refactor and make naming consistent and sensical
- [ ] need to test linux, should work though (probably)
- [ ] move script builds out of tree to avoid *.lib *.exp file littering
- [ ] maybe other things


SapphireMordred:
I really dislike the naming of actions and effects.
StatusEffect50 should instead be <Status<Name><Id>>
And same for Action. <Action<Name><Id>>
it becomes way more readable.
or subfolder so Status/Action can be removed from the script itself.

SapphireMordred:
Any reason why player is being passed by value in script calls?
eg.:
```cpp
 virtual void onTalk(uint32_t eventId, Core::Entity::Player player, uint64_t actorId) { }
```

GokuWeedLord:
The scripts in scripts/native/* can be called whatever you like, they're just mimicking the class names

there's no particular reason that player is passed by value, it's just what was used inside ScriptManager and I copied the prototypes almost verbatim - should be playerptr

SapphireMordred:
should be a const ref if you're not planning to change it, or ref if it should be editable.
name should be const ref.

src/servers/Server_Zone/Script/NativeScript.cpp
```diff
@@ -0,0 +1,139 @@
...
+        ZoneScript* NativeScript::getZoneScript( uint32_t zoneId )
+        {
+           auto script = m_zoneScripts.find( zoneId );
+           if( script == m_zoneScripts.end() )
+              return nullptr;
+
+           return script->second;
+        }
+
+        bool NativeScript::loadScript( std::string path )
```
itsMaru: any reason why this isn't pass by reference + const? from what I can tell you don't need mutability for path

src/servers/Server_Zone/Script/NativeScript.cpp
```diff
@@ -0,0 +1,139 @@
...
+        {
+           auto info = m_loader.loadModule( path );
+           if( info )
+           {
+              // todo: this is shit
+              if( auto script = m_loader.getScriptObject< StatusEffectScript >( info->handle, "StatusEffectScript" ) )
```
itsMaru: you're aware of this possible improvement but you're better off making a helper method "hasValidExports", passing the latter and having a switch case. 
Also "StatusEffectScript", "ActionScript" etc -> enum it. so you can just use hasValidExports( ScriptType scriptType )

src/servers/Server_Zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -519,5 +514,76 @@ void Core::DebugCommandHandler::serverInfo( char * data, Entity::Player& player,
 
 void Core::DebugCommandHandler::unlockCharacter( char* data, Entity::Player& player, boost::shared_ptr< DebugCommand > command )
 {
-   player.unlock( );
+   player.unlock();
+}
+
+void Core::DebugCommandHandler::script( char* data, Entity::Player &player, boost::shared_ptr< DebugCommand > command )
```
itsMaru: this isn't really an issue with the PR per se as we're not doing this currently, but it'd be in my eyes a good code improvement:

in general for smart pointers we use an using directive to help with verbosity
try using DebugCommandPtr = boost::shared_ptr< DebugCommand >;
then you can use DebugCommandPtr in argument type

src/servers/Server_Zone/Script/NativeScript.cpp
```diff
@@ -0,0 +1,139 @@
+#include "NativeScript.h"
+
+namespace Core {
+    namespace Scripting {
```
itsMaru: style - no indentation on namespace here

src/servers/Server_Zone/Script/NativeScript.cpp
```diff
@@ -0,0 +1,139 @@
...
+        StatusEffectScript* NativeScript::getStatusEffectScript( uint32_t statusId )
+        {
+           auto script = m_statusEffectScripts.find( statusId );
+           if( script == m_statusEffectScripts.end() )
+              return nullptr;
+
+           return script->second;
+        }
+
+        ActionScript* NativeScript::getActionScript(uint32_t abilityId)
```
itsMaru: spaces (arguments)

src/servers/Server_Zone/Script/NativeScript.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+           return true;
+        }
+
+        const std::string NativeScript::getModuleExtension()
+        {
+           return m_loader.getModuleExtension();
+        }
+
+        bool NativeScript::unloadScript( std::string name )
```
itsMaru: any reason why this isn't pass by reference + const? from what I can tell you don't need mutability for name

src/servers/Server_Zone/Script/NativeScript.cpp
```diff
@@ -0,0 +1,139 @@
...
+        {
+           auto info = m_loader.getScriptInfo( name );
+           if( info )
+           {
+              auto ptr = info->script;
+
+              if( removeValueFromMap< uint32_t, StatusEffectScript* >( ptr, m_statusEffectScripts ) )
```
itsMaru: is there no better way to do this without this amount of branching?

src/servers/Server_Zone/Script/NativeScript.h
```diff
@@ -0,0 +1,64 @@
...
+#include <boost/shared_ptr.hpp>
+#include <boost/make_shared.hpp>
+#include <boost/filesystem.hpp>
+
+#include <Server_Common/Crypt/md5.h>
+
+#include "ScriptLoader.h"
+
+namespace Core {
+    namespace Scripting {
```
itsMaru: no indentation for namespace

src/servers/Server_Zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -31,6 +31,8 @@
 #include "Session.h"
 #include <boost/make_shared.hpp>
 
+#include "Script/NativeScript.h"
```
itsMaru: include order

src/servers/Server_Zone/Script/NativeScript.cpp
GokuWeedLord: fixed

src/servers/Server_Zone/Script/NativeScript.cpp
GokuWeedLord: fixed

src/servers/Server_Zone/Script/NativeScript.cpp
GokuWeedLord: not really lol, fixed too

src/servers/Server_Zone/DebugCommand/DebugCommandHandler.cpp
SapphireMordred: check out Forwards.h

bin/config/settings_rest.xml
```diff
@@ -6,7 +6,7 @@
       <!-- Ip the lobby server listens on -->
       <ListenIp>127.0.0.1</ListenIp>
       <!-- Path of FFXIV dat files -->
-      <DataPath>C:\\SquareEnix\\FINAL FANTASY XIV - A Realm Reborn\\game\\sqpack\\ffxiv</DataPath>
+      <DataPath>H:\\SteamLibrary\\steamapps\\common\\FINAL FANTASY XIV - A Realm Reborn\\game\\sqpack\\ffxiv</DataPath>
```
takhlaq: ew steam client, og best client

bin/config/settings_zone.xml
```diff
@@ -1,12 +1,11 @@
 <Settings>
-
    <General>
       <!-- Port the zone server accepts game conenctions on -->
       <ListenPort>54992</ListenPort> 
       <!-- Ip the zone server conenctions on -->
       <ListenIp>127.0.0.1</ListenIp>
       <!-- Path of FFXIV dat files -->
-      <DataPath>C:\\SquareEnix\\FINAL FANTASY XIV - A Realm Reborn\\game\\sqpack\\ffxiv</DataPath>
+      <DataPath>H:\\SteamLibrary\\steamapps\\common\\FINAL FANTASY XIV - A Realm Reborn\\game\\sqpack\\ffxiv</DataPath>
```
takhlaq: still ew

PR #202, opened 2017-12-14
automagically reload rebuilt scripts

GokuWeedLord:


bin/config/settings_zone.xml
```diff
@@ -14,6 +14,7 @@
 
       	<HotSwap>
       		<Enabled>1</Enabled>
+            <ScriptsDir>../scripts/native/</ScriptsDir>
```
SapphireMordred: indendation?

src/servers/Server_Zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -594,10 +594,10 @@ void Core::DebugCommandHandler::script( char* data, Entity::Player &player, boos
       if( subCommand == params )
          player.sendDebug( "Command failed: requires name of script to reload" );
       else
-         if( g_scriptMgr.getNativeScriptHandler().reloadScript( params ) )
-            player.sendDebug( "Reloaded '" + params + "' successfully" );
-         else
-            player.sendDebug( "Failed to reload '" + params + "'" );
+      {
+         g_scriptMgr.getNativeScriptHandler().reloadScript( params );
+         player.sendDebug( "Attempting to reload script: " + params );
```
SapphireMordred: Is there any feedback of whether this actually succeeded? 
reload should probably return a bool to indicate that.

src/servers/Server_Zone/Script/NativeScriptManager.h
```diff
@@ -39,10 +42,13 @@ namespace Scripting {
 
       bool loadScript( const std::string& path );
       bool unloadScript( const std::string& name );
-      bool reloadScript( const std::string& name );
+      void reloadScript( const std::string &name );
```
SapphireMordred: Style

src/servers/Server_Zone/Script/ScriptManager.cpp
```diff
@@ -42,24 +52,65 @@ bool Core::Scripting::ScriptManager::init()
...
+
 void Core::Scripting::ScriptManager::loadDir( std::string dirname, std::set<std::string>& files, std::string ext )
 {
 
    g_log.info( "ScriptEngine: loading scripts from " + dirname );
 
    boost::filesystem::path targetDir( dirname );
 
-   boost::filesystem::recursive_directory_iterator iter( targetDir ), eod;
+   boost::filesystem::directory_iterator iter( targetDir ), eod;
```
SapphireMordred: We do not do multiple declaration per line, this is not documented anywhere, should be though ;)

src/servers/Server_Zone/DebugCommand/DebugCommandHandler.cpp
GokuWeedLord: reloading happens on the next server tick to avoid a rare condition where the lock on `/cache/<script>.dll` still exists and can't be overwritten so you can't really return from it anymore with useful information

it does show up in the server log if it is successful or not though. was debating on whether it would be a good idea to sendDebug to anyone logged in with a gmlvl > 0 that reloads were successful or not but I was thinking the server log would suffice

SapphireMordred: So it is not a reload call but rather a queueing of a reload process, this should probably be indicated by the name to reduce ambiguity.

GokuWeedLord: sounds good, will change that

PR #211, opened 2017-12-30
Ported the Quarrels with Squirrels Quest from Chai

AcedArmy:
Ported the Quarrels with Squirrels Quest from the old ChaiScript over to Native.
I cannot Test the Port up to Killing a NPC right now since a Bug or some other Reason in Sapphire, but after that is fixed it should work fine.


scripts/native/quest/subquest/gridania/SubFst002.cpp
```diff
@@ -0,0 +1,76 @@
...
+   {
+      auto callback = [&]( Entity::Player& player, uint32_t eventId, uint16_t param1, uint16_t param2, uint16_t param3 )
+      {
+         if( param2 == 1 ) // accept quest
+         {
+            player.updateQuest( getId(), SEQ_1 );
+         }
+      };
+
+      player.eventPlay ( getId (), 0, NONE, callback );
```
SapphireMordred: Strange space you got there ;)

PR #213, opened 2018-01-05
Ported the Quarrels with Squirrels Quest from Chai

AcedArmy:
Ported the Quarrels with Squirrels Quest from the old ChaiScript over to Native.
I cannot Test the Port up to Killing a NPC right now since a Bug or some other Reason in Sapphire, but after that is fixed it should work fine.


scripts/native/quest/subquest/gridania/SubFst002.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+      player.eventPlay( getId (), 1, NONE, callback );
+   }
+
+public:
+   SubFst002() : EventScript( "SubFst002", 65561 ) {}
+
+   void onTalk( uint32_t eventId, Entity::Player& player, uint64_t actorId ) override
+   {
+      auto actor = Event::mapEventActorToRealActor( static_cast< uint32_t >(actorId) );
```
itsMaru: space in func call arguments
static_cast< uint32_t >(actorId) -> static_cast< uint32_t >( actorId )

PR #234, opened 2018-01-31
zone, api: Removed old EXD reader

goaaats:
This will need some testing, and some things in ex.json might want to be changed for consistency:

* Unify modifier fields in ClassJob, TribeInfo, ParamGrow
* Rename ParamGrow->baseSpeed to something more fitting
* Action->preservesCombo was called is_instant in our old data, and seemingly still has that function
* Rename various fields in TerritoryType, Map, Level, Aetheryte to be more descriptive

This would have to be looked into with Clorifex to not break things that depend on SaintCoinach.


src/servers/sapphire_zone/Zone/Zone.cpp
```diff
@@ -225,7 +225,28 @@ uint8_t Zone::getNextWeather()
...
+
+   auto weatherRate = g_exdDataGen.getWeatherRate( weatherRateNum );
+   auto weatherRateFields = g_exdDataGen.m_WeatherRateDat.get_row( weatherRateNum );
+
+   std::map< uint8_t, int32_t> weatherRateMap;
+
+   uint8_t sumPc = 0;
+   for( size_t i = 0; i < 16; )
+   {
+      int32_t weatherId = boost::get<int32_t>( weatherRateFields[i] );
```
SapphireMordred: style

src/servers/sapphire_zone/Zone/Zone.cpp
```diff
@@ -225,7 +225,28 @@ uint8_t Zone::getNextWeather()
...
+
+      if( weatherId == 0 )
+         break;
+
+      sumPc += boost::get<uint8_t>( weatherRateFields[i + 1] );
```
SapphireMordred: style

src/servers/sapphire_zone/Action/ActionCast.cpp
```diff
@@ -31,7 +31,7 @@ Core::Action::ActionCast::ActionCast( Entity::ActorPtr pActor, Entity::ActorPtr
    m_startTime = 0;
    m_id = actionId;
    m_handleActionType = HandleActionType::Spell;
-   m_castTime = g_exdData.getActionInfo( actionId )->cast_time; // TODO: Add security checks.
+   m_castTime = g_exdDataGen.getAction( actionId )->cast100ms; // TODO: Add security checks.
```
SapphireMordred: whenever this is used, it now needs to be *100 iirc as that is how it was originally stored as discussed earlier

PR #247, opened 2018-02-08
hierachyId dump + scripts structure cleanup

GokuWeedLord:


src/servers/sapphire_zone/Zone/InstanceContent.cpp
```diff
@@ -31,6 +31,7 @@ Core::InstanceContent::InstanceContent( boost::shared_ptr< Core::Data::InstanceC
      m_instanceContentId( instanceContentId ),
      m_state( Created )
 {
+   m_instanceExpireTime = Util::getTimeSeconds() + ( m_instanceContentInfo->timeLimitmin * 60u );
```
SapphireMordred: this is not correct, expire time should not be set until the duty actually starts.

PR #251, opened 2018-02-10
objectcontrol packet implementation & debug commands

GokuWeedLord:


src/servers/sapphire_zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -768,6 +768,45 @@ void Core::DebugCommandHandler::instance( char* data, Entity::Player &player, bo
 
       instance->setVar( static_cast< uint8_t >( index ), static_cast< uint8_t >( value ) );
    }
+   else if( subCommand == "objupdate" )
+   {
+      uint32_t instanceId;
```
itsMaru: can't we just grab instance ID from the player? seems like it'd get uuuh repetitive

src/servers/sapphire_zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -768,6 +768,45 @@ void Core::DebugCommandHandler::instance( char* data, Entity::Player &player, bo
...
+      uint32_t objId;
+
+      sscanf( params.c_str(), "%d %d", &instanceId, &objId );
+
+      auto pInstance = g_territoryMgr.getInstanceZonePtr( instanceId );
+      if( !pInstance )
```
itsMaru: I feel like adding error messages to each of these silent returns would be nice, to see exactly what went wrong (those debug commands could be used just for checking if something's working as intended for example). what do you think?

src/servers/sapphire_zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -768,6 +768,45 @@ void Core::DebugCommandHandler::instance( char* data, Entity::Player &player, bo
...
+      if( !obj )
+         return;
+
+      instance->updateInstanceObj( obj );
+   }
+   else if( subCommand == "objstate" )
+   {
+      uint32_t instanceId;
+      uint32_t objId;
+      uint32_t state;
```
itsMaru: these types seem mismatched (iirc state was 8bit?). would be neat to match their sizes properly. if you do so though, look below

src/servers/sapphire_zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -768,6 +768,45 @@ void Core::DebugCommandHandler::instance( char* data, Entity::Player &player, bo
...
+
+      sscanf( params.c_str(), "%d %d %d", &instanceId, &objId, &state );
```
itsMaru: if you're matching the types don't forget format specifiers here

src/servers/sapphire_zone/Zone/InstanceContent.cpp
```diff
@@ -222,6 +222,8 @@ void Core::InstanceContent::registerInstanceObj( Core::Entity::InstanceObjectPtr
    object->setParentInstance( InstanceContentPtr( this ) );
 
    m_instanceObjects[object->getId()] = object;
+
+   g_log.debug("Registered instance eobj: " + std::to_string( object->getId() ) );
```
itsMaru: ÿ

src/servers/sapphire_zone/Actor/InstanceObject.cpp
```diff
@@ -33,7 +34,7 @@ void Core::Entity::InstanceObject::setState( uint8_t state )
 {
    m_state = state;
 
-   m_parentInstance->updateInstanceObj( InstanceObjectPtr( this ) );
+   //m_parentInstance->updateInstanceObj( InstanceObjectPtr( this ) );
```
itsMaru: intentional?

src/servers/sapphire_zone/DebugCommand/DebugCommandHandler.cpp
GokuWeedLord: yeah I thought about this but I don't think its worthwhile, mainly because these are likely to change soon anyway given that the current stuff is very wip

src/servers/sapphire_zone/DebugCommand/DebugCommandHandler.cpp
GokuWeedLord: sscanf expects `unsigned short` if you use `%hu` which `uint8_t` isnt so it throws a warning or whatever so I did it like this instead

itsMaru: %hhu

src/servers/sapphire_zone/Zone/InstanceContent.cpp
GokuWeedLord: i didn't know if it worked or not and attaching a debugger is too big brained for me :(

src/servers/sapphire_zone/Actor/InstanceObject.cpp
GokuWeedLord: yeah it crashed and i couldn't be bothered to figure out why lmao

GokuWeedLord: need to change how it works anyway cause i'm not a fan of the cyclic dependency anyway

src/servers/sapphire_zone/DebugCommand/DebugCommandHandler.cpp
```diff
@@ -768,6 +768,45 @@ void Core::DebugCommandHandler::instance( char* data, Entity::Player &player, bo
...
+
+      auto pInstance = g_territoryMgr.getInstanceZonePtr( instanceId );
+      if( !pInstance )
+         return;
+
+      auto instance = boost::dynamic_pointer_cast< InstanceContent >( pInstance );
```
SapphireMordred: Should probably add a "getAsInstanceContent" into the baseclass to be consistent with what we do for gameobjects

src/servers/sapphire_zone/Zone/InstanceContent.cpp
```diff
@@ -238,8 +240,24 @@ void Core::InstanceContent::updateInstanceObj( Core::Entity::InstanceObjectPtr o
...
+
+      // ????
+      //eobjStatePacket.data().levelId = 4236873;
+      //eobjStatePacket.data().unknown2 = 5;
+      //eobjStatePacket.data().unknown1C = 1065353216;
+      //eobjStatePacket.data().unknown20 = 2147423605;
+      //eobjStatePacket.data().actorId = 1074105831;
+      //eobjStatePacket.data().unknown = 1;
+
+      playerIt.second->queuePacket( eobjStatePacket );
```
SapphireMordred: If we just add another object container to the zone and override the "spawn" in InstanceObject" we could just use existing in range logic for spawning and updating objects.

GokuWeedLord: yeah i'm just wondering whats the best way to store objects in Zone properly so we can have multiple copies of things in the event of gathering points and etc

multiple players on the same gathering point and etc, not sure about how we would handle this with the current implementation in InstanceContent

SapphireMordred: gathering points are a whole other can of worms. They have a different type too, so they do not even qualify as eventObject.

src/servers/sapphire_zone/DebugCommand/DebugCommandHandler.cpp
takhlaq: huhu

PR #254, opened 2018-02-25
opcodes, ontalk handling in instances, build fix (maybe)

GokuWeedLord:


src/servers/sapphire_zone/Actor/Actor.cpp
```diff
@@ -313,6 +313,12 @@ void Core::Entity::Actor::setCurrentZone( ZonePtr currZone )
    m_pCurrentZone = currZone;
 }
 
+/*! \return InstanceContentPtr to the current instance, null if not an instance or not set */
+Core::InstanceContentPtr Core::Entity::Actor::getCurrentInstance() const
+{
+   return getCurrentZone()->getAsInstanceContent();
```
takhlaq: hmm wouldn't we want to check if they're loaded into a zone trying to access it

SapphireMordred: I don't disagree. Casting into instanceContent without knowing it actually is one may be problematic.

GokuWeedLord: dynamic_pointer_cast checks the type first, if it doesn't match it'll just return nullptr

GokuWeedLord: rather, if the types are incompatible it'll be nullptr

takhlaq: looks like my phone ate some words, i meant check their current zone isnt actually nullptr before trying to call getAsInstance()

GokuWeedLord: ah yeah, that could be a problem if it's used elsewhere, will fix

PR #266, opened 2018-03-06
minor instances fixes

NotAdam:


src/servers/sapphire_zone/Zone/InstanceContent.cpp
```diff
@@ -119,6 +129,7 @@ void Core::InstanceContent::onUpdate( uint32_t currTime )
                                            getDirectorId(), 0x40000001, m_instanceContentInfo->timeLimitmin * 60u ) );
          }
 
+         m_pEntranceEObj->setState( 7 );
```
SapphireMordred: What if the instance has no entrance object loaded?

src/servers/sapphire_zone/Zone/InstanceContent.h
```diff
@@ -22,6 +22,9 @@ class InstanceContent : public Event::Director, public Zone
       DutyFinished
    };
 
+   /*! number of milliseconds after all players are ready for the instance to commence (spawn circle removed) */
+   const uint32_t INSTANCE_COMMENCE_DELAY = 1250;
```
SapphireMordred: We should only use upper case in defines.

PR #267, opened 2018-03-07
Fix some weirdness with cotire

ShelbyZ:
- cotire having unresolved externals with sapphire_zone/scripts build
- /MP was already enabled, so remove the duplicate
- still need to figure out what about pch causes unresolved symbol __


itsMaru:
conflicts

NotAdam:
I think i mentioned somewhere that setting `/MP` like this won't work because for whatever reason `/MP` never makes it to the target settings

SapphireMordred:
closing for now, please reopen as needed.

PR #286, opened 2018-04-18
Added a whole lot of GM Stuff

AcedArmy:
- Added the GM Icons and a few others I was messing with to the OnlineStatusMask. (there should be a better way to do it but it should be good for now)

- Added the GM Invis command. I most likely did this in the most shit way possible but non the less it works.

- Added GM Chat Types. These where only added for the existing types. (shout, yell, say). If the player is a GM rank of any kind it will override the existing type with the GM variant.

### DIFF
--- a/src/servers/sapphire_api/PlayerMinimal.h
+++ b/src/servers/sapphire_api/PlayerMinimal.h
@@ -150,6 +150,16 @@ namespace Core {
          m_gmRank = rank;
       }
 
+      bool getGminvis() const
+      {
+         return m_gmInvis;
+      }
+
+      bool setGminvis( bool invis )
+      {
+         m_gmInvis = invis;
+      }
+
       void createInvDbContainer( uint16_t slot ) const;
 
       uint32_t m_modelEquip[10];
@@ -177,6 +187,7 @@ namespace Core {
       uint8_t m_look[26];
 
       uint8_t m_gmRank;
+      bool m_gmInvis;
 
       char m_name[34];
 
--- a/src/servers/sapphire_api/PlayerMinimal.h
+++ b/src/servers/sapphire_api/PlayerMinimal.h
@@ -150,12 +150,12 @@ namespace Core {
          m_gmRank = rank;
       }
 
-      bool getGminvis() const
+      bool getGmInvis() const
       {
          return m_gmInvis;
       }
 
-      bool setGminvis( bool invis )
+      bool setGmInvis( bool invis )
       {
          m_gmInvis = invis;
       }
--- a/src/servers/sapphire_zone/Actor/Player.cpp
+++ b/src/servers/sapphire_zone/Actor/Player.cpp
@@ -1605,7 +1605,7 @@ void Core::Entity::Player::sendZoneInPackets( uint32_t param1, uint32_t param2 =
    auto zoneInPacket = ActorControlPacket143( getId(), ZoneIn, param1, param2, param3, param4 );
    auto SetStatusPacket = ActorControlPacket142( getId(), SetStatus, static_cast< uint8_t >( Entity::Chara::ActorStatus::Idle ) );
 
-   if( getGmInvis() )
+   if( !getGmInvis() )
       sendToInRangeSet( zoneInPacket, true );
       if( shouldSetStatus )
          sendToInRangeSet( SetStatusPacket );
--- a/src/servers/sapphire_zone/Actor/Player.cpp
+++ b/src/servers/sapphire_zone/Actor/Player.cpp
@@ -176,7 +176,7 @@ Core::Common::OnlineStatus Core::Entity::Player::getOnlineStatus()
    uint32_t statusDisplayOrder = 0xFF14;
    uint32_t applicableStatus = 0;
 
-   for( uint32_t i = 0; i < 64; i++ )
+   for( uint32_t i = 0; i < std::numeric_limits< uint64_t >::digits; i++ )
    {
       bool bit = ( m_onlineStatus >> i ) & 1;
 
--- a/src/servers/sapphire_zone/Actor/Player.cpp
+++ b/src/servers/sapphire_zone/Actor/Player.cpp
@@ -1600,18 +1600,18 @@ void Core::Entity::Player::sendTitleList()
    queuePacket( titleListPacket );
 }
 
-void Core::Entity::Player::sendZoneInPackets( uint32_t param1, uint32_t param2 = 0, uint32_t param3 = 0, uint32_t param4 = 0, bool pSetStatus = false )
+void Core::Entity::Player::sendZoneInPackets( uint32_t param1, uint32_t param2 = 0, uint32_t param3 = 0, uint32_t param4 = 0, bool shouldSetStatus = false )
 {
    auto zoneInPacket = ActorControlPacket143( getId(), ZoneIn, param1, param2, param3, param4 );
    auto SetStatusPacket = ActorControlPacket142( getId(), SetStatus, static_cast< uint8_t >( Entity::Chara::ActorStatus::Idle ) );
 
    if( getGmInvis() )
       sendToInRangeSet( zoneInPacket, true );
-      if( pSetStatus )
+      if( shouldSetStatus )
          sendToInRangeSet( SetStatusPacket );
    else
       queuePacket( zoneInPacket );
-      if (pSetStatus)
+      if ( shouldSetStatus )
          queuePacket( SetStatusPacket );
 
    setZoningType( Common::ZoneingType::None );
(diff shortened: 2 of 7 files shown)


NotAdam:
there's a much better way to do status flags, Mordred and I discussed it briefly before: 
col 2 inside onlinestatus.exd is the display priority, meaning lower values overwrite higher ones and what you've currently got isn't retail accurate behavior

as goat said, invis should just be done with displayflags, no need for another var

what's the difference between normal chat vs gm chat? I assume the only difference is that you can't hide the GM variants using chat filtering, which shouldn't be always on

src/servers/sapphire_zone/Actor/Player.cpp
```diff
@@ -1606,11 +1647,17 @@ void Core::Entity::Player::finishZoning()
    switch( getZoningType() )
    {
       case ZoneingType::None:
-         sendToInRangeSet( ActorControlPacket143( getId(), ZoneIn, 0x01 ), true );
+         if (getGmInvis() == false)
```
NotAdam: better way to do this would be to have a check before the switch statement and check displayflags for invisible and then just return if its set, although this has some issues too

might need to rewrite this a bit so GMs can always see GMs without having to duplicate checks everywhere like you've done here :p

src/servers/sapphire_zone/Actor/Player.cpp
```diff
@@ -1606,11 +1647,17 @@ void Core::Entity::Player::finishZoning()
...
+            sendToInRangeSet( ActorControlPacket143( getId(), ZoneIn, 0x01 ), true );
+         else
+            queuePacket( ActorControlPacket143( getId(), ZoneIn, 0x01 ) );
          break;
 
       case ZoneingType::Teleport:
-         sendToInRangeSet( ActorControlPacket143( getId(), ZoneIn, 0x01, 0, 0, 110 ), true );
+         if( getGmInvis() == false )
```
NotAdam: on another note, better to just do `!getGmInvis()` imo

AcedArmy:
```as goat said, invis should just be done with displayflags, no need for another var```
Invis was done with the displayflags, the bool was so the server could keep track if the GM had toggled it. (If that's what you meant)

```I assume the only difference is that you can't hide the GM variants using chat filtering, which shouldn't be always on```
Yes that's is the only difference. From what I have seen from retail though it seems GM's always have it on without a toggle for it. I could change that behavior so it checks if the person has a GM OnlineStatus if you would like that?

File: src/servers/sapphire_zone/Actor/Player.cpp (hunk 1 above)
Comment (NotAdam) [reply src/servers/sapphire_zone/Actor/Player.cpp]: ah I kinda half wrote what I was thinking and left out the juicy stuff

what I was getting at is that duplicating all this is pretty bad, would be better to rewrite this so the packet code isn't duplicated 4 times along with the possibility for GMs to see invis GMs too

Comment (SapphireMordred) [reply src/servers/sapphire_zone/Actor/Player.cpp]: Like it has already been stated, i am very hesitant to merge this as it should be implemented using the priorities.

File: src/servers/sapphire_zone/Network/Handlers/GMCommandHandlers.cpp [hunk 3]
```diff
@@ -224,6 +224,32 @@ void Core::Network::GameConnection::gm1Handler( const Packets::GamePacket& inPac
...
+         auto inRange = targetPlayer->getInRangeActors();
+         for( auto actor : inRange )
+         {
+            targetPlayer->despawn( actor->getAsPlayer() );
+            targetPlayer->spawn( actor->getAsPlayer() );
+         }
+         player.sendNotice( "Invisibility flag for " + targetPlayer->getName() + " was toggled to off" );
+         break;
+      }
+   }
```
SapphireMordred: This block feels like a hack i am sorry to say.

src/servers/sapphire_api/PlayerMinimal.h
```diff
@@ -150,6 +150,16 @@ namespace Core {
          m_gmRank = rank;
       }
 
+      bool getGminvis() const
```
SapphireMordred: Style

src/servers/sapphire_api/PlayerMinimal.h
```diff
@@ -150,6 +150,16 @@ namespace Core {
...
+      {
+         return m_gmInvis;
+      }
+
+      bool setGminvis( bool invis )
```
SapphireMordred: Style

src/servers/sapphire_zone/Actor/Player.cpp
```diff
@@ -167,6 +177,16 @@ Core::Common::OnlineStatus Core::Entity::Player::getOnlineStatus()
...
    uint64_t rpMask = uint64_t( 1 ) << static_cast< uint32_t >( OnlineStatus::Roleplaying );
 
+   uint64_t prodMask = uint64_t( 1 ) << static_cast< uint32_t >( OnlineStatus::Producer );
+   uint64_t gmMask = uint64_t( 1 ) << static_cast< uint32_t >( OnlineStatus::GameMaster );
+   uint64_t gm1Mask = uint64_t( 1 ) << static_cast< uint32_t >( OnlineStatus::GameMaster1 );
+   uint64_t gm2Mask = uint64_t( 1 ) << static_cast< uint32_t >( OnlineStatus::GameMaster2 );
+
+   uint64_t menMask = uint64_t( 1 ) << static_cast< uint32_t >( OnlineStatus::Mentor );
+   uint64_t retMask = uint64_t( 1 ) << static_cast< uint32_t >( OnlineStatus::Returner );
+   uint64_t trialMask = uint64_t( 1 ) << static_cast< uint32_t >( OnlineStatus::TrialAdventurer );
```
takhlaq: should we not just `constexpr` these

NotAdam: nah, we should be getting them from the exd and then picking the one with the highest display priority (lowest value)

src/servers/sapphire_zone/Actor/Player.cpp
```diff
@@ -1642,22 +1600,34 @@ void Core::Entity::Player::sendTitleList()
    queuePacket( titleListPacket );
 }
 
+void Core::Entity::Player::sendZoneInPackets( uint32_t param1, uint32_t param2 = 0, uint32_t param3 = 0, uint32_t param4 = 0, bool pSetStatus = false )
```
NotAdam: the bool isnt a pointer, should just be `setStatus` or `shouldSetStatus`

src/servers/sapphire_zone/Actor/Player.cpp
```diff
@@ -1642,22 +1600,34 @@ void Core::Entity::Player::sendTitleList()
...
+   auto zoneInPacket = ActorControlPacket143( getId(), ZoneIn, param1, param2, param3, param4 );
+   auto SetStatusPacket = ActorControlPacket142( getId(), SetStatus, static_cast< uint8_t >( Entity::Chara::ActorStatus::Idle ) );
+
+   if( !Entity::Chara::Invisible )
+      sendToInRangeSet( zoneInPacket, true );
+      if( pSetStatus )
+         sendToInRangeSet( SetStatusPacket );
+   else
+      queuePacket( zoneInPacket );
+      if (pSetStatus)
```
NotAdam: style

src/servers/sapphire_zone/Actor/Player.cpp
```diff
@@ -1642,22 +1600,34 @@ void Core::Entity::Player::sendTitleList()
    queuePacket( titleListPacket );
 }
 
+void Core::Entity::Player::sendZoneInPackets( uint32_t param1, uint32_t param2 = 0, uint32_t param3 = 0, uint32_t param4 = 0, bool pSetStatus = false )
+{
+   auto zoneInPacket = ActorControlPacket143( getId(), ZoneIn, param1, param2, param3, param4 );
+   auto SetStatusPacket = ActorControlPacket142( getId(), SetStatus, static_cast< uint8_t >( Entity::Chara::ActorStatus::Idle ) );
+
+   if( !Entity::Chara::Invisible )
```
NotAdam: magic?

src/servers/sapphire_zone/Actor/Player.cpp
```diff
@@ -159,43 +169,32 @@ bool Core::Entity::Player::isMarkedForRemoval() const
...
+      return OnlineStatus::Online;
 
-   if( m_onlineStatus & meldMask )
-      status = OnlineStatus::LookingtoMeldMateria;
+   uint32_t statusDisplayOrder = 0xFF14;
+   uint32_t applicableStatus = 0;
 
-   if( m_onlineStatus & ptMask )
-      status = OnlineStatus::LookingforParty;
+   for( uint32_t i = 0; i < 64; i++ )
```
SapphireMordred: should not assume a static number here. That is bound to explode.

NotAdam: `for( uint32_t i = 0; i < std::numeric_limits< uint64_t >::digits; i++ )` pls

PR #287, opened 2018-04-21
Fixed making new zones and done some more gm stuff

AcedArmy:
- Fixed making new zones, and also made it look the way it does on createDefaultTerritories

- Added GuId to //gm teri_info and also fixed the spacing

- GM Raise is now using the new zoneinpacket system so it does not leak packets if invis

- GM Chat is now done via OnlineStatus since Adam complained about it being all the time :p

### DIFF
--- a/src/servers/sapphire_zone/Network/Handlers/GMCommandHandlers.cpp
+++ b/src/servers/sapphire_zone/Network/Handlers/GMCommandHandlers.cpp
@@ -538,10 +538,8 @@ void Core::Network::GameConnection::gm2Handler( const Packets::GamePacket& inPac
       targetPlayer->resetHp();
       targetPlayer->resetMp();
       targetPlayer->setStatus( Entity::Chara::ActorStatus::Idle );
+      targetPlayer->sendZoneInPackets( 0x01, 0x01, 0, 113, true );
 
-      targetPlayer->sendToInRangeSet( ActorControlPacket143( player.getId(), ZoneIn, 0x01, 0x01, 0, 113 ), true );
-      targetPlayer->sendToInRangeSet( ActorControlPacket142( player.getId(), SetStatus,
-         static_cast< uint8_t >( Entity::Chara::ActorStatus::Idle ) ), true );
       player.sendNotice( "Raised  " + targetPlayer->getName() );
       break;
    }
--- a/src/servers/sapphire_zone/Network/Handlers/GMCommandHandlers.cpp
+++ b/src/servers/sapphire_zone/Network/Handlers/GMCommandHandlers.cpp
@@ -467,10 +467,11 @@ void Core::Network::GameConnection::gm1Handler( const Packets::GamePacket& inPac
       auto pCurrentZone = player.getCurrentZone();
       player.sendNotice( "ZoneId: " + std::to_string( player.getZoneId() ) + "\nName: " +
                          pCurrentZone->getName() + "\nInternalName: " +
-                         pCurrentZone->getInternalName() + "\nPopCount: " +
+                         pCurrentZone->getInternalName() + "\nGuId: " +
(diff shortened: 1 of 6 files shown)


src/servers/sapphire_zone/Zone/TerritoryMgr.cpp
```diff
@@ -161,8 +161,9 @@ Core::ZonePtr Core::TerritoryMgr::createTerritoryInstance( uint32_t territoryTyp
    auto pZone = make_Zone( territoryTypeId, getNextInstanceId(), pTeri->name, pPlaceName->name );
    pZone->init();
 
-   m_territoryInstanceMap[pZone->getTerritoryId()][pZone->getGuId()] = pZone;
    m_instanceIdToZonePtrMap[pZone->getGuId()] = pZone;
+   m_territoryInstanceMap[pZone->getTerritoryId()][pZone->getGuId()] = pZone;
+   m_zoneSet.insert( { pZone } );
```
SapphireMordred: If i am not mistaken, this would result in zonelogic being processed twice.
instance and zone are handled seperatly.

NotAdam: nah, `m_instanceZoneSet` is responsible for the instance update loop whereas `m_zoneSet` is for normal zones for the same purpose. `m_territoryInstanceMap` is just so you can get all instances of a specific territory, not sure if we actually use that for anything at the moment though

src/servers/sapphire_zone/Network/Handlers/GMCommandHandlers.cpp
```diff
@@ -467,10 +467,11 @@ void Core::Network::GameConnection::gm1Handler( const Packets::GamePacket& inPac
       auto pCurrentZone = player.getCurrentZone();
       player.sendNotice( "ZoneId: " + std::to_string( player.getZoneId() ) + "\nName: " +
                          pCurrentZone->getName() + "\nInternalName: " +
-                         pCurrentZone->getInternalName() + "\nPopCount: " +
+                         pCurrentZone->getInternalName() + "\nGuId: " +
```
NotAdam: would be nice if you could fix this so the label comes before the getter on each line

    "\nName: " + pCurrentZone->getName() +
    "\nInternalName: " + pCurrentZone->getInternalName() +
    "\nGuid: " + .... etc

src/servers/sapphire_zone/Network/Handlers/PacketHandlers.cpp
```diff
@@ -515,23 +515,23 @@ void Core::Network::GameConnection::chatHandler( const Packets::GamePacket& inPa
    {
    case ChatType::Say:
    {
-      if (player.getGmRank() > 0)
+      if (player.getOnlineStatus() == OnlineStatus::GameMaster || player.getOnlineStatus() == OnlineStatus::GameMaster1 || player.getOnlineStatus() == OnlineStatus::GameMaster2 )
```
NotAdam: honestly this probably would be better to have another method inside player which returns true if these are set, cause no doubt there's going to be other stuff down the line that we're gonna want to switch based on whether a player is acting as a gm or not, and the duplication here isnt that great anyway 😛

src/servers/sapphire_zone/Zone/TerritoryMgr.cpp
SapphireMordred: Right, brainfart and ambigious naming.

NotAdam: yeah, naming isn't very good here, perhaps rename it to `m_zoneUpdateSet` and `m_instanceUpdateSet`?

NotAdam: not sure if that's any better though lmao

PR #291, opened 2018-05-13
ManSea001 now work properly

DantestyleXD:
Quest Script: ManSea001_00107
Quest Name: Coming to Limsa Lominsa

Old script:
- Events locked at Scene00003 (no QuestUpdate)

Actually:
- QuestUpdate set on Scene00003
- Add "If param2" on Scene00006 (because Actor1 ask if we want change position)

PS: I modified the entire script for a better understanding (imho)

### DIFF
--- a/src/servers/Scripts/quest/ManSea001.cpp
+++ b/src/servers/Scripts/quest/ManSea001.cpp
@@ -119,8 +119,8 @@ private:
       {
          if( result.param2 == 1 ) // finish quest
          {
-            if( player.giveQuestRewards( getId(), 0 ) )
-               player.finishQuest( m_id );
+            if (player.giveQuestRewards( getId(), 0))
+               player.finishQuest( getId() );
          }
       } );
    }
--- a/src/servers/Scripts/quest/ManSea001.cpp
+++ b/src/servers/Scripts/quest/ManSea001.cpp
@@ -1,13 +1,18 @@
 #include <Script/NativeScriptApi.h>
 #include <Actor/Player.h>
 #include "Event/EventHelper.h"
-#include "../ScriptObject.h"
+#include <ScriptObject.h>
+
+// Quest Script: ManSea001_00107
+// Quest Name: Coming to Limsa Lominsa
+// Quest ID: 65643
+// Start NPC: 1001028
+// End NPC: 1002697
 
 class ManSea001 : public EventScript
 {
 private:
-   // Steps in this quest ( 0 is before accepting,
-   // 1 is first, 255 means ready for turning it in
+
    static constexpr auto SEQ_0 = 0;
    static constexpr auto SEQ_1 = 1;
    static constexpr auto SEQ_FINISH = 255;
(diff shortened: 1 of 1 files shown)


src/servers/Scripts/quest/ManSea001.cpp
```diff
@@ -27,102 +32,98 @@ class ManSea001 : public EventScript
 
    void Scene00000( Entity::Player& player )
    {
-      auto callback = [this]( Entity::Player& player, const Event::SceneResult& result )
+      player.playScene( m_id, 0, 8192,
+         [&]( Entity::Player& player, const Event::SceneResult& result )
       {
-         if( result.param2 == 1 )
+         if ( result.param2 == 1 )
```
SapphireMordred: No space after control statement

src/servers/Scripts/quest/ManSea001.cpp
```diff
@@ -27,102 +32,98 @@ class ManSea001 : public EventScript
...
 
    void Scene00003( Entity::Player& player )
    {
-      auto callback = [this]( Entity::Player& player, const Event::SceneResult& result )
-      {
-         player.playScene( OPENING_EVENT_HANDLER, 0x1E, HIDE_HOTBAR | NO_DEFAULT_CAMERA, 1, 0 );
-      };
-
-      player.playScene( getId(), 3, NONE, 0, 0, callback );
+      player.playScene( OPENING_EVENT_HANDLER, 0x1E, HIDE_HOTBAR | NO_DEFAULT_CAMERA, 1, 0 );
```
SapphireMordred: This is a bit weird, because playscene0003 is supposed to play scene 3, you skip scene 3 and go straight to 0x1E

src/servers/Scripts/quest/ManSea001.cpp
```diff
@@ -27,102 +32,98 @@ class ManSea001 : public EventScript
...
-      };
-
-      player.playScene( getId(), 0, HIDE_HOTBAR, 0, 0, callback );
+      } );
    }
 
    void Scene00001( Entity::Player& player )
    {
-      player.playSceneChain( getId(), 1, DISABLE_SKIP | HIDE_HOTBAR | SET_BASE, bindScene( &ManSea001::Scene00002 ) );
+      player.playSceneChain( m_id, 1, DISABLE_SKIP | HIDE_HOTBAR | SET_BASE, bindScene( &ManSea001::Scene00002 ) );
```
NotAdam: anything that needs the script id should call `getId` not use `m_id` because we can change the behavior of the getter easily but not the variable

src/servers/Scripts/quest/ManSea001.cpp
DantestyleXD: The result is the same cause Scene00003 is an empty scene (according to the LUA file) and OPENING_EVENT_HANDLER works with the playSceneChain from Scene00002

src/servers/Scripts/quest/ManSea001.cpp
DantestyleXD: Every playSceneChain and playScene need getId() or only playSceneChain? (In the first case, I need to modify a lot of scripts xD)

takhlaq: ctrl shift h
m_id, 
getId(),

src/servers/Scripts/quest/ManSea001.cpp
SapphireMordred: If that is the case, do not call it Scene00003 if it really is playing 0x1E ( 30 ) since it really isn't playing 3. Otherwise it gets quite confusing

DantestyleXD: That call Scene00030 from OpeningLimsa (Scene00030 is not present into cpp autogenerated)

SapphireMordred: Fact is still that the call Scene0003 make no sense if it is not actually playing scene 3. It logically is flawed. If all it does is call to the opening handler, then sure ( should be named accordingly ) although i bet retail still uses scene3 as dummy to call to the openinghandler just as in gridania opening.

DantestyleXD: Check on retail:
Scene00002 call Scene00003 -> Scene00003 call opening.Scene00030

PR #292, opened 2018-05-15
ManSea002 now work properly

DantestyleXD:
Before:
- No start cutscene
- Wrong settings

Now:
- Start cutscene
- All quest credits work
- MSQ can be complete

NOTE: For a properly zoning, player need to accept Mansea002 and Subsea001.

### DIFF
--- a/src/servers/Scripts/quest/ManSea002.cpp
+++ b/src/servers/Scripts/quest/ManSea002.cpp
@@ -35,23 +35,26 @@ private:
 
    void Scene00000( Entity::Player& player )
    {
-      auto callback = [&]( Entity::Player& player, const Event::SceneResult& result )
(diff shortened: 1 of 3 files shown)


SapphireMordred:
``(when EOBJ system will be implement, I will adjust this one).``
That is actually possible already. Gridania has such a collect quest implemented ( at the amphitheater iirc )

SapphireMordred:
Here is an example of such a quest in chaiscript
https://github.com/SapphireMordred/Sapphire/blob/04e7f1c9cb461e4179bbd311c31fff0a68a47daf/scripts/chai/quest/SubFst015.chai

DantestyleXD:
ManSea002/ManSea003 and SubSea001 are complete.
Let me know if there are something that you don't like :smile:

PR #294, opened 2018-05-19
Return (Action 6) now work (again?)

DantestyleXD:
Just Return (action 6) now teleport correctly player to the favourite Aetheryte


NotAdam:
Just as a note before you get a bit overzealous adding more actions, we are planning to rewrite the action system soon:tm: so while this is appreciated, don't waste your time too much at the moment because I don't know how much it's going to change

PR #316, opened 2018-06-11
Some adjust for future extractions (quest_parse)

DantestyleXD:


src/servers/sapphire_zone/Script/NativeScriptApi.h
```diff
@@ -88,7 +88,7 @@ class EventScript : public ScriptObject
       ScriptObject( questId, typeid( EventScript ).hash_code() )
    { }
 
-   virtual void onTalk( uint32_t eventId, Entity::Player& player, uint64_t actorId ) { }
+   virtual void onTalk( uint32_t eventId, Entity::Player& player, uint32_t actorId ) { }
```
NotAdam: the client sends us a 64bit uint with the actor id, you shouldn't be truncating it just because it doesn't make a difference at the moment

src/tools/quest_parser/main.cpp
```diff
@@ -24,9 +24,9 @@ Core::Logger g_log;
 Core::Data::ExdDataGenerated g_exdDataGen;
 
 const std::string onTalkStr(
-   "   void onTalk( uint32_t eventId, Entity::Player& player, uint64_t actorId ) override\n"
+   "   void onTalk( uint32_t eventId, Entity::Player& player, uint32_t actorId ) override\n"
    "   {\n"
-   "      auto actor = Event::mapEventActorToRealActor( actorId );\n"
+   "      uint32_t actor = Event::mapEventActorToRealActor( actorId );\n"
```
NotAdam: Why are you changing it from `auto` to `uint32_t`?

src/tools/quest_parser/main.cpp
```diff
@@ -67,6 +67,9 @@ void createScript( boost::shared_ptr< Core::Data::Quest >& pQuestData, std::set<
       "// In order for this script to be loaded, change its extension to .cpp\n"
       "// and move it to the correct folder in <root>/scripts/native/\n"
       "\n"
+      "#include <Script/NativeScriptApi.h>\n"
```
NotAdam: `NativeScriptApi.h` is already included by `ScriptObject.h`

src/tools/quest_parser/main.cpp
```diff
@@ -147,7 +150,7 @@ void createScript( boost::shared_ptr< Core::Data::Quest >& pQuestData, std::set<
 
    if( !pQuestData->itemReward0.empty() )
    {
-      rewards += "      static constexpr auto RewardItem[] = { ";
+      rewards += "      uint16_t RewardItem[] = { ";
```
NotAdam: We have the type inferred by the compiler because if we ever change what we output, we don't have to manage the type there too - there's no real benefits to explicitly declaring the type

src/tools/quest_parser/main.cpp
```diff
@@ -258,7 +261,7 @@ void createScript( boost::shared_ptr< Core::Data::Quest >& pQuestData, std::set<
    for( auto enemy : enemy_ids )
    {
       scriptEntry += std::string(
-         "   void onMobKill_" + std::to_string( enemy ) + "( Entity::Player& player )\n"
+         "   void onMobKill_" + std::to_string( enemy ) + "( Entity::Player& player, uint64_t npcId )\n"
```
NotAdam: This is incorrect too, the function is `onNpcKill( uint32_t npcId, Entity::Player& player )`

You'll also need to indicate which NPC IDs are to be handled either by outputting a comment in the body of the generated function or by generating a switch/if statement

src/servers/sapphire_zone/Script/NativeScriptApi.h
DantestyleXD: At the moment, VS consider it like a warning and I didn't found problems ingame (atm)

NotAdam: auto actor = static_cast< uint32_t >( actorId );

src/tools/quest_parser/main.cpp
DantestyleXD: 'cause auto -> uint64_t and VS doesn't like it

PR #323, opened 2018-06-23
common, zone: Streamline ActorControlTypes, refactor ActionHandler/ClientTrigger to ClientActorControl

goaaats:


SapphireMordred:
I'd appreciate if we stuck to ClientTrigger instead of ClientActorControlType for now.
If that is changed i will merge this.

src/common/Network/PacketDef/Ipcs.h
```diff
@@ -227,7 +227,7 @@ namespace Packets {
       ReqCountdownCancel       = 0x012D, // updated 4.2
 
       ZoneLineHandler          = 0x013C, // updated 4.3
-      ActionHandler            = 0x013D, // updated 4.3
+      ClientActorControl            = 0x013D, // updated 4.3
```
SapphireMordred: did a tab sneak in there?

src/servers/sapphire_zone/Network/Handlers/ClientTriggerHandler.cpp
```diff
@@ -131,7 +62,7 @@ void Core::Network::GameConnection::actionHandler( const Packets::GamePacket& in
 
    switch( commandId )
    {
-       case ClientTrigger::ToggleSeathe:  // Toggle sheathe
+       case ClientActorControlType::ToggleSeathe:  // Toggle sheathe
```
SapphireMordred: All these should be changed back too.

PR #331, opened 2018-07-03
segment header types, send in range packets and override their target id

NotAdam:


src/common/Network/PacketContainer.cpp
```diff
@@ -52,9 +53,16 @@ void Core::Network::Packets::PacketContainer::fillSendBuffer( std::vector< uint8
...
    for( ; it != m_entryList.end(); ++it )
    {
-      auto data = (*it)->getData();
-      memcpy( &tempBuffer[0] + sizeof( FFXIVARR_PACKET_HEADER ) + offset, &data[0], (*it)->getSize() );
-      offset += (*it)->getSize();
+      auto pPacket = (*it);
+
+      if( pPacket->getSegmentType() == SEGMENTTYPE_IPC )
+      {
+         pPacket->setTargetActor( m_segmentTargetOverride );
```
SapphireMordred: may only want to do this if m_segmentTargetOverride != 0

PR #337, opened 2018-07-08
Classquest Archer

DantestyleXD:
This include all Classquests for Archer:

- Way of the Archer
- A Matter of Perspective
- School of Hard Nocks
- Violators Will Be Shot
- To Catch a Poacher
- Homecoming
- The One That Got Away

--------------------------------------------------------------------
Issue

- [ ] Duty calls (QuestBattles)
- [ ] Battle Bnpc system
- [x] Eobj doesn't want despawn (I really don't understand why...) (ClsArc004)
- [ ] Hunting Log #338 
- [ ] Crash on "EventActionStart" if perform a Player Skill on Eobj #339


src/servers/Scripts/quest/classquest/ARC/ClsArc000.cpp
```diff
@@ -0,0 +1,180 @@
...
+      }
+      if( actor == Actor1 && player.getQuestSeq( getId() ) == SeqFinish )
+      {
+         Scene00002( player );
+      }
+   }
+
+   void onMobKill( Entity::Player& player, uint64_t npcId )
+   {
+      if( npcId != Enemy0 || Enemy1 || Enemy2 )
```
SapphireMordred: i assume you actually want to check if npcId is != any of those, in that case you have to write 
if( npcId != Enemy0 || npcId != Enemy1 || npcId != Enemy2 )

src/servers/Scripts/quest/classquest/ARC/ClsArc000.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+      if( QUEST_ONKILL_37 == 1 && QUEST_ONKILL_49 == 1 && QUEST_ONKILL_47 == 1 )
+      {
+         player.updateQuest( questId, SeqFinish );
+      }
+   }
+
+   void Scene00000( Entity::Player& player )
+   {
+      player.playScene( getId() , 0, HIDE_HOTBAR,
```
SapphireMordred: style ( remove space before , "

src/servers/Scripts/quest/classquest/ARC/ClsArc000.cpp
```diff
@@ -0,0 +1,180 @@
...
+            if( result.param2 == 1 )
+            {
+               player.updateQuest( getId(), Seq1 );
+            }
+         } );
+   }
+
+   void Scene00001( Entity::Player& player )
+   {
+      player.playScene( getId() , 1, SET_EOBJ_BASE | HIDE_HOTBAR | INVIS_EOBJ,
```
SapphireMordred: same thing, and a couple after this.

src/servers/Scripts/quest/classquest/ARC/ClsArc001.cpp
```diff
@@ -0,0 +1,470 @@
...
+            {
+               Scene00008( player );
+            },
+            nullptr, eventId );
+      }
+   }
+
+   void onMobKill( Entity::Player& player, uint64_t npcId )
+   {
+      if( npcId != Enemy0 || Enemy1 )
```
SapphireMordred: As previously mentioned, npcId != missing

src/servers/Scripts/quest/classquest/ARC/ClsArc000.cpp
```diff
@@ -0,0 +1,180 @@
...
+   }
+
+   void onMobKill( Entity::Player& player, uint64_t npcId )
+   {
+      if( npcId != Enemy0 || Enemy1 || Enemy2 )
+         return;
+
+      auto currentKC37 = player.getQuestUI8AL( getId() ) + 1;
+      auto currentKC49 = player.getQuestUI8BH( getId() ) + 1;
+      auto currentKC47 = player.getQuestUI8BL( getId() ) + 1;
```
NotAdam: this isn't correct, you increment each counter regardless of which mob was killed

DantestyleXD: QuestBits are inside the lua script
```
    if L6_41 == A0_35.SEQ_2 then
      if A3_38 == A0_35.ENEMY0 then
        L7_42 = A1_36:GetQuestUI8AL(L5_40)
        if L7_42 >= 3 then
          return false
        end
        return 3 > A1_36:GetQuestUI8AL(L5_40)
      elseif A3_38 == A0_35.ENEMY1 then
        L7_42 = A1_36:GetQuestUI8BH(L5_40)
        if L7_42 >= 3 then
          return false
        end
        return 3 > A1_36:GetQuestUI8BH(L5_40)
      elseif A3_38 == A0_35.ENEMY2 then
        L7_42 = A1_36:GetQuestUI8BL(L5_40)
        if L7_42 >= 3 then
          return false
        end
        return 3 > A1_36:GetQuestUI8BL(L5_40)
      end
    end
```

src/servers/Scripts/quest/classquest/ARC/ClsArc000.cpp
waterlgndx: The code you suggested isn't correct and would always be true it should be :
if( npcId != Enemy0 && npcId != Enemy1 && npcId != Enemy2 )
         return;

src/servers/Scripts/quest/classquest/ARC/ClsArc000.cpp
NotAdam: which isn't what you're doing, you're incrementing the kill count for each one regardless of the npc killed

you also don't check the quest sequence either

src/servers/Scripts/quest/classquest/ARC/ClsArc000.cpp
```diff
@@ -73,9 +73,9 @@ class ClsArc000 : public EventScript
       if( npcId != Enemy0 || npcId != Enemy1 || npcId != Enemy2 )
```
waterlgndx: Unless I'm missing something, the code suggested won't work.  As it is written now this if statement would always be true.  What you really need is something like:
```
 if( npcId != Enemy0 && npcId != Enemy1 && npcId != Enemy2 )
         return; 
```
or you could write:
```
if (!(npcId == Enemy0 || npcId == Enemy1 || npcId == Enemy2))
  return;
```

Issue #308, opened 2018-06-11
Class/job changes when equipping items


NotAdam:
Player class should change automagically when equipping items that are restricted to another class that is currently unlocked

PR #402, opened 2018-10-18
Add class change on weapon swaps. Also adds a forgotten inventory on character creation.

Squall5668:
Before merging, please keep in mind that this will crash the zone server if you try to login with a corrupt character (specifically one that is holding a weapon of a different class).

Referencing #308 for progress.


SapphireMordred:
We should add handling for said crash scenario. E.g. remove the equipped item if not applicable.

Squall5668:
Removing a characters weapon looks even more broken than the current behavior. This commit will, instead, keep the current behavior. Corrupt characters can now logon without crashing zone server.

Like equipItem, equipWeapon now has an bool in order to know if it can send an update. It will be carried over to it. I also think that equipWeapon could be called from equipItem instead of updateModels, but it might cause some duplicate code, so I'll leave it as is for now.

PR #437, opened 2018-11-23
saving player names in map to reduce SQL query

XeAri:


src/servers/sapphire_zone/ServerMgr.h
```diff
@@ -59,6 +59,7 @@ class ServerMgr
 
   std::map< uint32_t, SessionPtr > m_sessionMapById;
   std::map< std::string, SessionPtr > m_sessionMapByName;
+  std::map< uint32_t, std::string > m_payerNameMap;
```
NotAdam: `m_playerNameMap`

SapphireMordred: Name should also indicate what is mapped to the name. See the other maps

NotAdam:
while this kinda alleviates the problem, it also means if the server is running for a long time and someone changes their name, it won't update on the house

src/servers/sapphire_zone/ServerMgr.cpp
```diff
@@ -380,11 +380,19 @@ std::string Core::ServerMgr::getPlayerNameFromDb( uint32_t playerId )
 
   std::string playerName = res->getString( 1 );
 
-  m_payerNameMap[ playerId ] = playerName;
+  m_payerNameMapById[ playerId ] = playerName;
```
SapphireMordred: Good, except i don't think you are looking for a payer name here ;)

PR #440, opened 2018-11-26
building mostly works now, houses are saved & restored on reboot

NotAdam:


src/servers/sapphire_zone/Zone/HousingZone.cpp
```diff
@@ -148,23 +163,26 @@ void Core::HousingZone::sendLandUpdate( uint8_t landId )
...
+
 
     if( auto house = pLand->getHouse() )
     {
-      // todo: this is retarded, need a getter to the internal array
-      for( int i = 0; i < 8; i++ )
-      {
-        auto slot = static_cast< Common::HousePartSlot >( i );
-        auto part = pLand->getHouse()->getHousePart( slot );
+      auto& parts = house->getHouseParts();
```
SapphireMordred: style

SapphireMordred: nvm

PR #445, opened 2018-12-01
spawn internal house instances

NotAdam:


src/servers/sapphire_zone/Territory/Housing/HousingInteriorTerritory.h
```diff
@@ -10,9 +11,14 @@ namespace Sapphire::World::Territory::Housing
...
+    virtual ~HousingInteriorTerritory();
+
     bool init() override;
 
     void onPlayerZoneIn( Entity::Player& player ) override;
     void onUpdate( uint32_t currTime ) override;
-  }
+
+  private:
+    uint64_t m_landIdent;
```
SapphireMordred: Might aswell use our LandIdent struct, if you need the uint 32 access, can always be used as union

PR #447, opened 2018-12-04
More housing and gilshop fix

NotAdam:


src/servers/sapphire_zone/Network/Handlers/ClientTriggerHandler.cpp
```diff
@@ -427,6 +427,28 @@ void Sapphire::Network::GameConnection::clientTriggerHandler( const Packets::FFX
 
       break;
     }
+    case ClientTriggerType::RequestEstateGreeting:
```
SapphireMordred: This should probably also be moved to housingMgr. Keep logic out of the trigger handler.

PR #451, opened 2018-12-10
Issue template

goaaats:


takhlaq:
can we have sapphire and client version as the top to save time on outdated/version mismatch issues

PR #466, opened 2018-12-31
Revert "Avoid Cmake that create C++14 build with C++17 installed"

NotAdam:
Reverts SapphireServer/Sapphire#461


SapphireMordred:
we'll do this differently.

PR #471, opened 2019-01-06
Several crashes fixed, GM Jump and Call fixed and closer to retail

### LINKED ISSUE #341: GM Jump/Call issues
GM Jumping to a character within an instance while being outside of said instance crashes the zone server. The same occurs if a character within an instance GM Calls a character that is outside of said instance. The issue occurs even if both characters are properly bound to the instance.

Also, a very minor issue I noticed, but since you guys want to be as close as possible to retail behavior, I figured I'd mention it:

On Sapphire, when GM Jumping/Calling a character that is in the same zone as the player using the command, it will not trigger a loading screen. However, on retail, I've been teleported by GMs several times, and even within the same zone, it always triggered a loading screen.

JeidoUran:
This should fix #341 , I did extensive tests and there were no crashes or odd behaviors anymore. That being said, some things I noted regarding GM Call and Jump:

I used `exitInstance` if the player is inside an instance but is it actually necessary? I'm not sure if it's automatically called whenever the player exits an instanced area by any means.

Also, ZoneIn packets seems to always gets sent twice. In this case, they get sent 4 times, so I assume `changePosition` already automatically initiates a ZoneIn, but the params are different from the ones I captured and added to the commands, so should I leave them in?


src/world/Network/Handlers/GMCommandHandlers.cpp
```diff
@@ -610,12 +622,26 @@ void Sapphire::Network::GameConnection::gm2Handler( FrameworkPtr pFw,
...
+      }
+      if( targetPlayer->getCurrentInstance() && player.getCurrentInstance() != targetPlayer->getCurrentInstance() )
+      {
+        auto instance = targetPlayer->getCurrentInstance();
+        auto pInstanceContent = instance->getAsInstanceContent();
+        // Not sure if GMs actually get bound to an instance they jump to on retail. It's mostly here to avoid a crash for now.
+        pInstanceContent->bindPlayer( player.getId() );
+        player.setInstance( targetPlayer->getCurrentInstance() );
+      }
       if( targetPlayer->getZoneId() != player.getZoneId() )
```
NotAdam: this won't work properly for housing zones (interior and exterior areas) because the zone ids are the same.

What you should do here is get the current zone and then check the guids and then move the player accordingly, that should then be portable across every territory type unless we add something similar to instance binding to another territory type

NotAdam:
```suggestion
      if( targetPlayer.getCurrentZone()->getGuid() != player.getCurrentZone()->getGuid() )
```
guess something like this?

NotAdam:
as for `exitInstance`, what it does is returns a player to their position before they entered the instance. `setInstance` will store their current territory, position and rotation and then when you call `exitInstance`, it'll return the player to that position.

I don't think the current way we handle exit/set instance works all that well though, because there's a lot of other 'instances' which work the same but your traditional instance (house interiors, quest battles) which kinda work but not really. I think its fine for now at least, but when we tackle instances properly I reckon we'll want to take another look at it

src/world/Network/Handlers/GMCommandHandlers.cpp
JeidoUran: Yeah, I actually completely forgot about housing zones and it was indeed causing various issues and crashes left and right.

I changed it to what you suggested and it seems to be working fine now, I think I tested all cases.

There is one last thing bothering me (but I'm being a bit nitpicky here), `prepareZoning` doesn't have the time to fully fade out when moving to a different zone. It works fine when moving within the same zone and has a 1 second fade out, but it gets interrupted if moving to a different zone/instance. Is there a way to delay `setInstance` by 1 second or to have it ran only when `prepareZoning` is done?

NotAdam: yeah there's queuedZoning which does that iirc, but I don't think it'll work with any zone you throw at it - I believe it just takes a territory type id

NotAdam: https://github.com/SapphireServer/Sapphire/blob/aea0bf333a3b0fb3282d4ac3d24ad411b1613045/src/world/Actor/Player.h#L18-L33

So that'd work fine, but would need to be changed to take a zoneptr. I think we'll need to go through and change most stuff to queued zoning anyway because as is, a lot of stuff will instantly move you and it ruins muh immersion

PR #473, opened 2019-01-06
move all configuration items to new config class

NotAdam:


config/global.ini.default
```diff
@@ -0,0 +1,25 @@
...
+Database = sapphire
+Username = sapphire
+Password =
+SyncThreads = 2
+AsyncThreads = 2
+
+[Parameters]
+ServerSecret = default
+DataPath = C:\\SquareEnix\\FINAL FANTASY XIV - A Realm Reborn\\game\\sqpack
+WorldID = 67
```
SapphireMordred: this should be in world

NotAdam: it's needed for lobby/world

SapphireMordred: Ah, fair enough. We shall revisit it if we decide to allow multiple worlds.

NotAdam: yeah, I only moved it into global because we using it heaps in housing and we've hardcoded it everywhere

NotAdam: speaking of which, going to rename this to general

PR #482, opened 2019-01-19
fix bnpcs not despawning

NotAdam:


src/world/Actor/BNpc.cpp
```diff
@@ -306,6 +311,10 @@ void Sapphire::Entity::BNpc::update( int64_t currTime )
 
     case BNpcState::Idle:
     {
+      // passive mobs should ignore players unless aggro'd
+      if( m_aggressionMode == 1 )
```
takhlaq: we should enum these at some point

src/world/Network/PacketWrappers/NpcSpawnPacket.h
```diff
@@ -76,17 +76,14 @@ namespace Sapphire::Network::Packets::Server
       m_data.bNPCName = bnpc.getBNpcNameId();
 
       m_data.state = 1;
```
takhlaq: this too

src/world/Actor/BNpc.cpp
takhlaq: in fact, i think there's already enums for these in exds

NotAdam: I can't find it in the exds, I know it was there at one point but looked before making the PR and couldn't find it

maybe i'm blind?

PR #501, opened 2019-01-23
Basic pathfinding

goaaats:
* Adds a new NaviMgr Manager class that loads NaviProviders which provide pathfinding services to NPCs in Zones
* Adds basic path tracking code to BNpc

### DIFF
--- a/src/world/Actor/BNpc.cpp
+++ b/src/world/Actor/BNpc.cpp
@@ -315,7 +315,7 @@ void Sapphire::Entity::BNpc::update( int64_t currTime )
 {
   const uint8_t minActorDistance = 4;
   const uint8_t aggroRange = 8;
-  const uint8_t maxDistanceToOrigin = 30;
+  const uint8_t maxDistanceToOrigin = 1000;
 
   if( m_status == ActorStatus::Dead )
     return;
--- a/src/world/Actor/BNpc.cpp
+++ b/src/world/Actor/BNpc.cpp
@@ -161,8 +161,8 @@ bool Sapphire::Entity::BNpc::moveTo( const FFXIVARR_POSITION3& pos )
     for(int i = 0; i < path.size(); i++)
       Logger::debug("{0}: {1} {2} {3}", i, path[i].x, path[i].y, path[i].z);
 
-    face( path[0] );
-    setPos(path[0]);
+    face( path[1] );
+    setPos(path[1]);
   }
 
   sendPositionUpdate();
--- a/src/world/Actor/BNpc.cpp
+++ b/src/world/Actor/BNpc.cpp
@@ -31,7 +31,9 @@
 #include "BNpcTemplate.h"
 #include "Manager/TerritoryMgr.h"
 #include "Common.h"
+#include "Framework.h"
 #include <Logging/Logger.h>
+#include <Manager/NaviMgr.h>
 
 using namespace Sapphire::Common;
 using namespace Sapphire::Network::Packets;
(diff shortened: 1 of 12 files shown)


src/world/Actor/BNpc.cpp
```diff
@@ -151,13 +154,85 @@ void Sapphire::Entity::BNpc::setState( BNpcState state )
...
+    // Targets are the same
+    return false;
+
+  // Check if we have to recalculate
+  if( Util::getTimeMs() - m_naviLastUpdate > 500 )
+  {
+    auto pNaviMgr = m_pFw->get< World::Manager::NaviMgr >();
+    auto pNaviProvider = pNaviMgr->getNaviProvider( m_pCurrentZone->getInternalName() );
+
+    if(!pNaviProvider)
```
SapphireMordred: style

src/world/Manager/DebugCommandMgr.cpp
```diff
@@ -367,6 +367,18 @@ void Sapphire::World::Manager::DebugCommandMgr::set( char* data, Entity::Player&
       }
     }
   }
+  else if( subCommand == "mobaggro" )
+  {
+    auto inRange = player.getInRangeActors();
+
+    for( auto actor : inRange )
+      {
```
SapphireMordred: indendation is off

src/world/Navi/NaviProvider.h
```diff
@@ -0,0 +1,72 @@
...
+    std::string m_internalName;
+
+    dtNavMesh* m_naviMesh;
+    dtNavMeshQuery* m_naviMeshQuery;
+
+    float m_polyFindRange[3];
+
+  private:
+    static int fixupCorridor( dtPolyRef* path, const int npath, const int maxPath,
+                          const dtPolyRef* visited, const int nvisited );
```
SapphireMordred: style

src/world/Navi/NaviProvider.h
```diff
@@ -0,0 +1,72 @@
...
+    static int fixupShortcuts( dtPolyRef* path, int npath, dtNavMeshQuery* navQuery );
+    inline static bool inRange( const float* v1, const float* v2, const float r, const float h );
+    static bool getSteerTarget( dtNavMeshQuery* navQuery, const float* startPos, const float* endPos,
```
SapphireMordred: indendation

src/world/Actor/BNpc.cpp
```diff
@@ -151,13 +154,85 @@ void Sapphire::Entity::BNpc::setState( BNpcState state )
...
+    {
+      Logger::error( "No NaviProvider for zone#{0} - {1}", m_pCurrentZone->getGuId(), m_pCurrentZone->getInternalName() );
+      return false;
```
takhlaq: for now maybe we should log an error but continue with pathing to random pos around actor instead of not moving at all if navmesh is unavailable

NotAdam: how would you path with no navmesh? just do the naive pathing we had before with the interp between bnpc/player?

takhlaq: ya

src/world/Actor/BNpc.cpp
```diff
@@ -151,13 +154,85 @@ void Sapphire::Entity::BNpc::setState( BNpcState state )
...
+    if( !path.empty() )
+    {
+      m_naviLastPath = path;
+      m_naviTarget = pos;
+      m_naviPathStep = 0;
+      m_naviLastUpdate = Util::getTimeMs();
+    }
+    else
+    {
+      Logger::debug( "No path found for target: {0} {1} {2} in ", pos.x, pos.y, pos.z, m_pCurrentZone->getInternalName() );
```
NotAdam: does this actually work? should need `{4}`?

src/world/Actor/BNpc.cpp
NotAdam: dunno, might just be better to scream and yell about there being no nav provider instead of using that

there seems to be an issue at least from when I was playing with it where bnpcs would bug out and go under the map with zones that have a large y difference

PR #515, opened 2019-01-30
GM Tells fixed

JeidoUran:


NotAdam:
Is this what retail sends for gm tells?

JeidoUran:
Yep, did a GM call today and compared a regular /tell with a GM /tell.

NotAdam:
Nice, good find

PR #517, opened 2019-02-01
added blind and deaf to bnpcs

XeAri:
needs to be checked

### DIFF
--- a/src/world/Actor/BNpc.h
+++ b/src/world/Actor/BNpc.h
@@ -98,9 +98,9 @@ namespace Sapphire::Entity
 
     void setBlind( bool state );
 
-    bool isDeaf();
+    bool getIsDeaf();
 
-    bool isBlind();
+    bool getIsBlind();
 
     void regainHp();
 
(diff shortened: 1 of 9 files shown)


src/world/Network/Handlers/PacketHandlers.cpp
```diff
@@ -256,6 +256,7 @@ void Sapphire::Network::GameConnection::updatePositionHandler( FrameworkPtr pFw,
     unk1 = 0x7F;
     unk2 = 0x00;
     unk4 = 0x3C;
+    player.setRunning( true );
```
NotAdam: we shouldn't be trusting the client about things like this

src/world/Actor/BNpc.h
```diff
@@ -94,6 +94,14 @@ namespace Sapphire::Entity
...
     void setTimeOfDeath( uint32_t timeOfDeath );
 
+    void setDeaf( bool state );
+
+    void setBlind( bool state );
+
+    bool isDeaf();
+
+    bool isBlind();
+
```
NotAdam: should group these like getIsDeaf/setIsDeaf & getIsBlind/setIsBlind

PR #522, opened 2019-02-06
Improved player momvement

XeAri:


src/common/Common.h
```diff
@@ -25,6 +25,8 @@ namespace Sapphire::Common
     float x;
     float y;
     float z;
+    inline bool operator == ( const FFXIVARR_POSITION3& target ) const
+    { return ( this->x == target.x && this->y == target.y && this->z == target.z ); }
```
NotAdam: I would move this into its own cpp file because it'd be nice to get the other operators in there too along with moving some of the vector math from util into the position structure for some more ease of use

src/world/Network/Handlers/PacketHandlers.cpp
```diff
@@ -239,73 +202,57 @@ void Sapphire::Network::GameConnection::updatePositionHandler( FrameworkPtr pFw,
...
-
-  uint8_t unk1 = 0;
-  uint8_t unk2 = 0;
-  uint8_t unk3 = moveState;
-  uint16_t unk4 = 0;
+  auto clientAnimationType = updatePositionPacket.data().clientAnimationType;
+  auto animationState = updatePositionPacket.data().animationState;
+  auto animationType = updatePositionPacket.data().animationType;
+  auto headRotation = updatePositionPacket.data().headPosition;
+  uint8_t unkownRotation = 0;
```
NotAdam: if we're not going to set this, just make it an optional param and ignore it in here

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -665,10 +665,11 @@ struct FFXIVIpcActorMove :
...
 {
   /* 0000 */ uint8_t rotation;
-  /* 0001 */ uint8_t unknown_1;
+  /* 0001 */ uint8_t headRotation;
   /* 0002 */ uint8_t animationType;
-  /* 0003 */ uint8_t unknown_3;
-  /* 0004 */ uint16_t unknown_4;
+  /* 0003 */ uint8_t animationState;
+  /* 0004 */ uint8_t animationSpeed;
+  /* 0005 */ uint8_t unkownRotation;
```
NotAdam: typo

src/world/Network/PacketWrappers/MoveActorPacket.h
```diff
@@ -19,21 +19,22 @@ namespace Sapphire::Network::Packets::Server
...
 
       m_data.rotation = Util::floatToUInt8Rot( actor.getRot() );
-      m_data.unknown_1 = unk1;
+      m_data.headRotation = headRotation;
       m_data.animationType = animationType;
-      m_data.unknown_3 = unk3;
-      m_data.unknown_4 = unk4;
+      m_data.animationState = state;
+      m_data.animationSpeed = animationSpeed;;
+      m_data.unkownRotation = unkownRotation;
```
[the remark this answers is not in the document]
XeAri: value seems like headRotation and its only set while you falling down. default value i noticed was 0x7F

src/world/Network/Handlers/PacketHandlers.cpp
XeAri: its already optional and only set while falling

NotAdam: ah missed that, np

src/common/Vector3.h
```diff
@@ -0,0 +1,18 @@
+#ifndef _VECTOR3_H_
+#define _VECTOR3_H_
+
+namespace Sapphire
+{
+  namespace Common
```
NotAdam: don't nest namespaces, do `Sapphire::Common` in the outer block

src/common/Vector3.cpp
```diff
@@ -0,0 +1,8 @@
+#include <string.h> 
+
+#include "Vector3.h"
+
+inline bool Sapphire::Common::FFXIVARR_POSITION3::operator == ( const FFXIVARR_POSITION3& target ) const
+{
+  return !memcmp( this, &target, sizeof( FFXIVARR_POSITION3 ) );
```
NotAdam: you can't compare floats like this due to floating point precision errors

PR #523, opened 2019-02-06
Mostly retail like movement + bnpcs auto attack move delay

XeAri:


src/world/Actor/BNpc.cpp
```diff
@@ -521,7 +522,12 @@ void Sapphire::Entity::BNpc::update( int64_t currTime )
         }
 
         if( distance > minActorDistance )
-          moveTo( pHatedActor->getPos() );
+        {
+          uint64_t tick = Util::getTimeMs();
```
NotAdam: currTime is the current time in ms, no need to get it again

src/world/Actor/BNpc.cpp
```diff
@@ -521,7 +522,12 @@ void Sapphire::Entity::BNpc::update( int64_t currTime )
...
+          auto pTeriMgr = m_pFw->get< World::Manager::TerritoryMgr >();
+          if ( ( tick - m_lastAttack ) > 600 && pTeriMgr->isDefaultTerritory( getCurrentZone()->getTerritoryTypeId() ) )
```
NotAdam: is there a particular reason for this check? not sure why this is needed

AriAvery: checked retail, every bnpcs stays a few after auto attack

NotAdam: no I mean the `pTeriMgr->isDefaultTerritory( getCurrentZone()->getTerritoryTypeId() )`

AriAvery: I'm not sure about raid bosses or dungeons

NotAdam: hmmm, alright, maybe leave it out for now then? not much point in doing stuff like this when we're not certain how it works

my guess is it's consistent, but would be better to know for sure before we start adding stuff like this in

AriAvery: alright

PR #525, opened 2019-02-10
Pugilist and Thaumaturge "Close to Home"

DanteMccloud:
Added and tested the "Close to Home" quests ManWil003 and ManWil004 for PGL and THM.


NotAdam:
Thanks for doing these, looks good. Can you PR this from develop though? don't want to merge these into master for now

PR #532, opened 2019-03-17
add missing zonepositions, use move semantics for framework constructor

NotAdam:


src/api/SapphireAPI.cpp
```diff
@@ -10,15 +10,8 @@
 
 #include <Database/DatabaseDef.h>
 
-Sapphire::Network::SapphireAPI::SapphireAPI()
```
SapphireMordred: just do this in the header, no need to declare default in cpp

PR #542, opened 2019-04-07
Action target selection

NotAdam:


src/world/Util/ActorFilter.h
```diff
@@ -29,18 +30,15 @@ namespace Sapphire::World::Util
...
-  //
+  /////////////////////////////////////////////////////////////////////////////
+
+  class ActorFilterSingleTarget : public ActorFilter
+  {
+    uint32_t m_actorId;
+  public:
+    explicit ActorFilterSingleTarget( uint32_t actorId );
+    bool conditionApplies( const Entity::Actor& actor ) override;
+  };
```
SapphireMordred: I don't think we want to filter when we already know the affected target. This would be a waste of time.

PR #547, opened 2019-04-16
ActorControl, OpCode and NpcSpawn findings. Also added worldId to Examine and PlayerSpawn

JeidoUran:
Adding the worldId to Examine fixed the World Name appearing as "Unknown" when examining a player.
Adding it to PlayerSpawn fixed the World Name not showing up in the HUD, and fixed right-clicking a player to send a /tell.

### DIFF
--- a/src/common/Network/CommonActorControl.h
+++ b/src/common/Network/CommonActorControl.h
@@ -77,7 +77,7 @@ enum ActorControlType : uint16_t
     TreasureScreenMsg = 0x57,
     SetOwnerId = 0x59,
     ItemRepairMsg = 0x5C,
-    BluActionLearn = 0x63, // Displays "Action Learned!" animation. Only used for visual, doesn't actually unlock it. param1: actionId. param2: 0 = text only, 1 = text+animation
+    BluActionLearn = 0x63, // Displays "Action Learned!" animation. Only used for visual, doesn't actually unlock it. param1: actionId. param2: 0 = text only, 1 = text+animation. Verification needed
 
     DirectorInit = 0x64,
     DirectorClear = 0x65,
--- a/src/common/Network/CommonActorControl.h
+++ b/src/common/Network/CommonActorControl.h
@@ -77,6 +77,7 @@ enum ActorControlType : uint16_t
     TreasureScreenMsg = 0x57,
     SetOwnerId = 0x59,
     ItemRepairMsg = 0x5C,
+    BluActionLearn = 0x63, // Triggers the "Action Learned!" animation and tells which action is learned. Any action ID may be used
 
     DirectorInit = 0x64,
     DirectorClear = 0x65,
@@ -287,6 +288,8 @@ enum ActorControlType : uint16_t
     SetDutyActionHud = 0x5E9, // disable/enable
     SetDutyActionActive = 0x5EA,
     SetDutyActionRemaining = 0x5EB,
+
+    EurekaStep = 0x73A, // alters the Eureka maps depending on the progress of the player
   };
 
   enum ClientTriggerType
--- a/src/common/Network/CommonActorControl.h
+++ b/src/common/Network/CommonActorControl.h
@@ -289,7 +289,7 @@ enum ActorControlType : uint16_t
     SetDutyActionActive = 0x5EA,
     SetDutyActionRemaining = 0x5EB,
 
-    EurekaStep = 0x73A, // alters the Eureka maps depending on the progress of the player
+    EurekaStep = 0x73A, // alters the progress of the player on Eureka (is used for all the eureka zones)
   };
 
   enum ClientTriggerType
--- a/src/common/Network/CommonActorControl.h
+++ b/src/common/Network/CommonActorControl.h
@@ -77,7 +77,7 @@ enum ActorControlType : uint16_t
     TreasureScreenMsg = 0x57,
     SetOwnerId = 0x59,
     ItemRepairMsg = 0x5C,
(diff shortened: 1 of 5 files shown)


src/world/Network/PacketWrappers/ExaminePacket.h
```diff
@@ -49,6 +49,8 @@ namespace Sapphire::Network::Packets::Server
         m_data.mainWeaponModel = pTarget->getModelMainWeapon();
         m_data.secWeaponModel = pTarget->getModelSubWeapon();
 
+        m_data.worldId = 67; // TODO: world id from server
```
goaaats: You should use the world ID variable from general config for these

src/world/Network/PacketWrappers/PlayerSpawnPacket.h
```diff
@@ -31,6 +31,10 @@ namespace Sapphire::Network::Packets::Server
       m_data.classJob = static_cast< uint8_t >( player.getClass() );
       //m_data.status = static_cast< uint8_t >( pPlayer->getStatus() );
 
+// TODO: world id from server
+      m_data.currentWorldId = 67;
```
goaaats: Same thing for those, please use the general config entry

src/world/Network/PacketWrappers/ExaminePacket.h
NotAdam: that won't really work unless you pull framework in which doesn't feel like a good solution in the case of packet wrappers

src/common/Network/PacketDef/Ipcs.h
```diff
@@ -182,6 +182,7 @@ namespace Sapphire::Network::Packets
     Mount = 0x01E8, // updated 4.5
 
     DirectorVars = 0x01EA, // updated 4.5
+    DirectorPopUp = 0x01F5, // display dialogue pop-ups in duties and FATEs, for example, Teraflare's countdown
```
NotAdam: `DirectorPopUp` isn't right I don't think, there's an exd with all of these in there from my knowledge so it should be called that, but in any case I don't think it has any relation to directors

NotAdam: ah yeah all in `InstanceContentTextData` so it could be related to directors but not sure, will need to suss out the packet

JeidoUran: I see. I thought it was related to directors because the packet I captured didn’t do anything unless it was injected after sending the correct DirectorInit for it.

src/world/Network/PacketWrappers/ExaminePacket.h
SapphireMordred: We should probably just store the current server in the player.

src/common/Network/PacketDef/Ipcs.h
SapphireMordred: What are the parameters used in this? Also i noticed that it may not be pulling just from that table but also from quest exd tables.

SapphireMordred: We've just looked into it, It really appears to be a DirectorMessage packet.

src/common/Network/CommonActorControl.h
```diff
@@ -77,6 +77,7 @@ enum ActorControlType : uint16_t
     TreasureScreenMsg = 0x57,
     SetOwnerId = 0x59,
     ItemRepairMsg = 0x5C,
+    BluActionLearn = 0x63, // Triggers the "Action Learned!" animation and tells which action is learned. Any action ID may be used
```
SapphireMordred: I wonder if this is exclusive to blue then. Might just be for any action/job?

JeidoUran: When BLU unlocks a spell, it sends a regular ToggleActionUnlock and 0x63 (0x63 is actually sent for each Blue Mages in the party since they all learn the spell at the same time, but oddly enough the animation isn't visible on other players) So I assume it's exclusive to BLU.

I also believe it's only used for visual purposes and doesn't actually unlock anything. I tried to inject the packet with an empty Blue Magic Spellbook and all it did was play the "Action Learned" animation and put the "newly acquired" green dot on the corresponding learned spell, but it didn't actually unlock it, even within the book.
blulearn
blulearn2

SapphireMordred: Be that as it may, as you've mentioned, you could use it on any action as a visual confirmation.
I find it unlikely ( granted not impossible ) that something with such a low id in the list would be exclusive to something so recently added.

SapphireMordred: I guess we can agree to put a comment saying //verification needed on it ;)

JeidoUran: Done! :P

I should have been clearer from the start, but this is the animation I was talking about:
blulearn3

However, you're totally right that the fact it's possible to only display the "You learn Song of Torment!" part without the big "Action Learned!", and that it accepts any actionId and not only BLU ones makes it usable for all jobs.

NotAdam: I think they might then use this for job quests? those show the little notification with "You learn x" don't they?

NotAdam: yeah pretty sure this is like `ActionLearnNotification` and `param2` is like the display variant or something. would need to check an old executable but don't have em at work, but my guess is that they've just added that to the same actor control as an additional param

NotAdam: the actor control opcode is also pretty low too which is indicative of the fact that it's probably been around for a while as well

PR #555, opened 2019-04-25
fix ilvl calc, somewhat accurate auto attack dmg, fix autoattack range

NotAdam:


src/world/Actor/Chara.cpp
```diff
@@ -729,4 +729,132 @@ Sapphire::Common::BaseParam Sapphire::Entity::Chara::getPrimaryStat() const
   assert( classJob );
 
   return static_cast< Sapphire::Common::BaseParam >( classJob->primaryStat );
+}
+
+uint32_t Sapphire::Entity::Chara::getStatValue( Sapphire::Common::BaseParam baseParam ) const
+{
+  uint32_t value = 0;
+
+  switch( baseParam )
```
SapphireMordred: All these ignore bonus stats from gear now if i read this right?

NotAdam: nah it adds the gear stat at the end

PR #561, opened 2019-05-28
fix self target actions not targeting correctly, fix chasing shadows being uncompletable

NotAdam:


src/scripts/instances/questbattles/ChasingShadows.cpp
```diff
@@ -67,6 +67,13 @@ class ChasingShadows : public Sapphire::ScriptAPI::QuestBattleScript
       return;
     }
 
+    if( instance.getCountEnemyBNpc() == 0 && successCalled == 0 )
```
Regiuz: If you have this before the spawn logic, you will complete the instance if you insta kill all currently spawned bnpcs, regardless if more are being spawned.

SapphireMordred: ^

NotAdam: yeah, changed it so we default to 0% hp if the boss stops existing which'll spawn all add phases instantly

PR #571, opened 2019-07-27
Update Ipcs.h

karashiiro:
Found this accidentally :v
I don't know how to code the packet information into the repository, but this is how I was processing it in JavaScript:
```
module.exports = (struct) => {
    let worldID = getWorld(struct.data[0x42]);
    struct.world = worldID ? worldID : struct.data[0x42];

    struct.fc = String.fromCodePoint(...struct.data.slice(0x105, 0x117)).replace(/\0/g, ""); // Why is this an odd number?
    struct.searchComment = String.fromCodePoint(...struct.data.slice(0x44, 0x80)).replace(/\0/g, "");

    struct.classJob = new Map();
    let classJobAbbreviation = ["GLD", "PGL", "MRD", "LNC", "ARC", "CNJ", "THM", "CRP", "BSM", "ARM",
    "GSM", "LTW", "WVR", "ALC", "CUL", "MIN", "BTN", "FSH", "CUL", "MNK", "WAR", "DRG", "BRD", "PLD", "BLM",
    "ACN", "SCH", "SMN", "ROG", "NIN", "MCH", "DRK", "AST", "SAM", "RDM", "BLU", "GNB", "DNC"];
    for (let i = 0; i < classJobAbbreviation.length; i++) {
        struct.classJob.set(classJobAbbreviation[i], struct.data[0x122 + i * 4]);
    }
};

const getWorld = (id) => { // WIP
    if (id ===  34) return "Brynhildr";
    if (id ===  35) return "Famfrit";
    if (id ===  37) return "Mateus";
    if (id ===  39) return "Omega";
    if (id ===  40) return "Jenova";
    if (id ===  41) return "Zalera";
    if (id ===  49) return "Kujata";
    if (id ===  53) return "Exodus";
    if (id ===  54) return "Faerie";
    if (id ===  55) return "Lamia";
    if (id ===  57) return "Siren";
    if (id ===  62) return "Diabolos";
    if (id ===  63) return "Gilgamesh";
    if (id ===  64) return "Leviathan";
    if (id ===  65) return "Midgardsormr";
    if (id ===  68) return "Atomos";
    if (id ===  71) return "Moogle";
    if (id ===  72) return "Tonberry";
    if (id ===  73) return "Adamantoise";
    if (id ===  74) return "Coeurl";
    if (id ===  75) return "Malboro";
    if (id ===  77) return "Ultros";
    if (id ===  78) return "Behemoth";
    if (id ===  80) return "Cerberus";
[...]

PR #572, opened 2019-07-27
Added ExamineSearchInfo opcode and packet structure, ID'd CP and GP in PlayerStats packet structure

karashiiro:
Correction of https://github.com/SapphireServer/Sapphire/pull/571.

I believe everything is in order, this time? I also renamed a couple `unknown` properties in PlayerStats to CP and GP, after a bit of extra testing.

Also, I'm not sure if this is intentional, but `Common::CLASSJOB_SLOTS` is 28, when there are 38 classes and jobs in total. https://github.com/SapphireServer/Sapphire/commit/b0eba69733425103e0d2fb4719a00c013686a0d0 has all of them and their IDs, but I collapsed them into an array in https://github.com/SapphireServer/Sapphire/commit/fbf82c937e991659d56782a8491a913c68249d80.

### DIFF
--- a/src/common/Network/PacketDef/Zone/ServerZoneDef.h
+++ b/src/common/Network/PacketDef/Zone/ServerZoneDef.h
@@ -135,7 +135,7 @@ namespace Sapphire::Network::Packets::Server
     char fcName[24];
     uint8_t unknown7;
     uint16_t padding1;
-    uint16_t classJobIdLevel[Common::CLASSJOB_SLOTS * 2]; // Pairs of classJobId, classJobLevel
+    uint16_t classJobIdLevel[76]; // Pairs of classJobId, classJobLevel (38 * 2)
     /*
     Like this:
     uint16_t gldId; // 1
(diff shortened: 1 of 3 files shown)


NotAdam:
`Common::CLASSJOB_SLOTS` is the number of exp/level slots because arr jobs use their base class entries instead

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -120,6 +120,30 @@ namespace Sapphire::Network::Packets::Server
...
+    uint32_t unknown3;
+    uint16_t unknown4;
+    uint16_t unknown5;
+    uint16_t unknown6;
+    uint8_t worldId;
+    char searchMessage[193];
+    char fcName[24];
+    uint8_t unknown7;
+    uint16_t padding1;
+    uint16_t classJobIdLevel[Common::CLASSJOB_SLOTS * 2]; // Pairs of classJobId, classJobLevel
```
NotAdam: would need to fix this otherwise size isn't going to be right, but what you could do is just a nested structure

```
struct ClassJobEntry
{
  uint16_t id;
  uint16_t level;
} levelEntries[38];
```
and you can add another constant to common for the number as well

karashiiro: oh, that's a better idea than what I just did, 1 sec :v

PR #573, opened 2019-08-07
added more FC opcodes and packet structures

karashiiro:


src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -247,6 +247,22 @@ namespace Sapphire::Network::Packets::Server
...
 
+  struct FFXIVIpcFreeCompanyMemberLogInOut : FFXIVIpcBasePacket< FreeCompanyMemberLogInOut >
+  {
+    uint16_t unknown;
+    uint16_t unknown1;
+    uint16_t unknown2;
+    uint16_t unknown3;
+    uint16_t unknown4;
+    char padding[6];
+    uint8_t status; // 0x0F Login, 0x10 Logout
```
NotAdam: do the other status codes do other things? this could be the same packet for the messages like executing company actions, people joining/leaving the fc, etc meaning that `FreeCompanyMemberLogInOut` isn't exactly correct

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -247,6 +247,22 @@ namespace Sapphire::Network::Packets::Server
...
+    uint8_t padding1;
+    char padding2[6];
+    uint16_t unknown5;
+    char fc[46];
```
NotAdam: fc name or? bit vague

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -247,6 +247,22 @@ namespace Sapphire::Network::Packets::Server
...
+    char character[32];
```
NotAdam: would probably make this `characterName` just so it's a bit more obvious

src/common/Network/PacketDef/Zone/ClientZoneDef.h
```diff
@@ -323,6 +323,16 @@ struct FFXIVIpcMarketBoardRequestItemListingInfo :
   /* 0000 */ uint32_t requestId;
 };
 
+struct FFXIVIpcFreeCompanyUpdateShortMessageHandler :
+  FFXIVIpcBasePacket< FreeCompanyUpdateShortMessageHandler >
+{
+  char shortMessage[104]; // Contains old/gibberish data in remaining slots if not filled
```
NotAdam: that's fine, the actual message is null terminated so you can ignore anything past it, gibberish or not

karashiiro:
I don't have a reliable method of testing if 12C is used for anything other than login/logout, but I have seen that join/leave events are handled with a different opcode (maybe it was FreeCompanyBoardMessage?). For now I've renamed it.

BiscuitUmU:
After some playing around with the `FreeCompanyEvent` packet, eventID changes to a whole bunch of messages (not sure where the link is yet)
image

I also feel that it should be moved to `src/common/Network/PacketDef/Chat/ServerChatDef.h`
since it is a chat packet.

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -247,20 +247,20 @@ namespace Sapphire::Network::Packets::Server
...
     char padding[6];
-    uint8_t status; // 0x0F Login, 0x10 Logout
+    uint8_t eventID; // 0x0F Login, 0x10 Logout
     uint8_t padding1;
     char padding2[6];
     uint16_t unknown5;
-    char fc[46];
-    char character[32];
+    char fcName[46];
+    char characterName[32];
```
BiscuitUmU: Based on testing they don't always seem to be fcName and characterName.
For example: `The company tag has been changed to "XXX"`. XXX uses what you put down as fcName.
So it's probably safe to assume they are parameters and don't have a set name.

karashiiro:
Oh, guess I was wrong about join/leave then, lots of interesting stuff there. Moving to `ServerChatDef.h` and renaming properties to generics. The specifics of what each field becomes can be added to the comments underneath for documentation. Is it possible for you to add what you've found so far? Fork my branch and chain PR or something, I haven't had to do that yet so I don't know how :v

NotAdam:
looks like eventid uses stuff from logmessage and fcactivity, the eventids might be mapped in the exe somewhere given that login/logout collides with others inside fcactivity

PR #578, opened 2019-08-16
MarketBoardItemListing updated opcode, reordered ServerZoneDef market structs

karashiiro:
The structs in ServerZoneDef didn't match Ipcs.h and everything else did :v There was also a typo in there.

As for the MarketBoardItemListing struct, the reason for the nested `unknown2` struct being looped is because it always contains the same data 1-3 times.

### DIFF
--- a/src/common/Network/PacketDef/Ipcs.h
+++ b/src/common/Network/PacketDef/Ipcs.h
@@ -106,7 +106,7 @@ namespace Sapphire::Network::Packets
     ReqMoogleMailLetter = 0x0139, // updated 5.0
     MailLetterNotification = 0x013A, // updated 5.0
 
-    MarketBoardItemListingCount = 0x0125, // updated 4.5
+    MarketBoardItemListingCount = 0x013B, // updated 5.0
     MarketBoardItemListing = 0x013C, // updated 5.0
     MarketBoardItemListingHistory = 0x012A, // updated 4.5
     MarketBoardSearchResult = 0x0139, // updated 4.5
--- a/src/common/Network/PacketDef/Ipcs.h
+++ b/src/common/Network/PacketDef/Ipcs.h
@@ -107,7 +107,7 @@ namespace Sapphire::Network::Packets
     MailLetterNotification = 0x013A, // updated 5.0
 
     MarketBoardItemListingCount = 0x0125, // updated 4.5
-    MarketBoardItemListing = 0x0126, // updated 4.5
+    MarketBoardItemListing = 0x013C, // updated 5.0
     MarketBoardItemListingHistory = 0x012A, // updated 4.5
     MarketBoardSearchResult = 0x0139, // updated 4.5
 
--- a/src/common/Network/PacketDef/Zone/ServerZoneDef.h
+++ b/src/common/Network/PacketDef/Zone/ServerZoneDef.h
@@ -313,11 +313,20 @@ namespace Sapphire::Network::Packets::Server
       uint32_t unknown9;
       uint32_t unknown10;
       uint32_t unknown11;
-      char padding2[16];
+      char unknown12[16]; // Materia? Doesn't seem to match any IDs
       char retainerName[64];
       bool hq;
-      uint16_t padding3;
-      uint8_t unknown12;
+      uint8_t materiaCount;
+      uint8_t padding3;
+      uint8_t retainerCity;
+      /**
+       * 0x01 Limsa Lominsa
+       * 0x02 Gridania
+       * 0x03 Ul'dah
+       * 0x04 Ishgard
+       * 0x07 Kugane
+       * 0x0A Crystarium
+       */
       uint64_t padding4;
     } listing[10]; // Multiple packets are sent if there are more than 10 search results.
     uint32_t unknown13;
--- a/src/common/Network/PacketDef/Zone/ServerZoneDef.h
+++ b/src/common/Network/PacketDef/Zone/ServerZoneDef.h
@@ -313,7 +313,7 @@ namespace Sapphire::Network::Packets::Server
       uint16_t unknown8;
       uint16_t unknown9;
       uint32_t unknown10;
-      uint16_t unknown11; // Always 30000
+      uint16_t durability;
       uint16_t padding3;
       uint16_t materiaValue[5];
       /**
--- a/src/common/Network/PacketDef/Zone/ServerZoneDef.h
+++ b/src/common/Network/PacketDef/Zone/ServerZoneDef.h
@@ -294,7 +294,7 @@ namespace Sapphire::Network::Packets::Server
 
   struct FFXIVIpcMarketBoardItemListing : FFXIVIpcBasePacket< MarketBoardItemListing >
   {
-    struct itemListing // 152 bytes each
+    struct ItemListing // 152 bytes each
     {
       uint32_t unknown; // Changes if multiple packets are sent for a single item
       uint8_t padding;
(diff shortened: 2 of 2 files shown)


src/common/Network/PacketDef/Ipcs.h
```diff
@@ -107,7 +107,7 @@ namespace Sapphire::Network::Packets
     MailLetterNotification = 0x013A, // updated 5.0
 
     MarketBoardItemListingCount = 0x0125, // updated 4.5
-    MarketBoardItemListing = 0x0126, // updated 4.5
+    MarketBoardItemListing = 0x013C, // updated 5.0
```
NotAdam: `MarketBoardItemListingCount` is probably 13b and `MarketBoardItemListingHistory` is would be 13d given they're in the same group

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -283,6 +283,100 @@ namespace Sapphire::Network::Packets::Server
...
+
+  struct FFXIVIpcMarketBoardItemListing : FFXIVIpcBasePacket< MarketBoardItemListing >
+  {
+    struct itemListing // 152 bytes each
+    {
+      uint32_t unknown; // Changes if multiple packets are sent for a single item
+      uint8_t padding;
+      uint16_t unknown1;
+      uint8_t padding1;
+      struct unknown2
```
NotAdam: is there a particular reason that you have this defined as a structure?

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -283,6 +283,100 @@ namespace Sapphire::Network::Packets::Server
...
+      uint32_t pricePerUnit;
+      uint32_t unknown7;
+      uint32_t itemQuantity;
+      uint32_t itemID;
+      uint16_t unknown8;
+      uint16_t unknown9;
+      uint32_t unknown10;
+      uint16_t unknown11; // Always 30000
+      uint16_t padding3;
+      uint16_t materiaValue[5]; // Materia ID = (materiaValue & 0xFF0) >> 4
```
NotAdam: might wanna put full docs here instead

```
    auto materiaId = (i & 0xFF0) >> 4;
    auto index = i & 0xF;
```

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -283,6 +283,100 @@ namespace Sapphire::Network::Packets::Server
...
+        uint16_t unknown5;
+      } unknown6[3];
+      uint32_t pricePerUnit;
+      uint32_t unknown7;
+      uint32_t itemQuantity;
+      uint32_t itemID;
+      uint16_t unknown8;
+      uint16_t unknown9;
+      uint32_t unknown10;
+      uint16_t unknown11; // Always 30000
```
NotAdam: durability, 30000 is 100%, 60000 is 200%

market might reset durability to 100% for buyers?

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -283,6 +283,100 @@ namespace Sapphire::Network::Packets::Server
...
+       * 0x0A Crystarium
+       */
+      uint64_t padding6;
+    } listing[10]; // Multiple packets are sent if there are more than 10 search results.
+    uint32_t unknown12;
+    char padding7[16];
+    uint8_t unknown13;
+    uint16_t padding8;
+    uint8_t unknown14;
+    uint64_t padding9;
```
NotAdam: not sure what this is, but there would be a u64 for the content id of a character so you can request the signature on an item, might be this, might be elsewhere in the packet

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -283,6 +283,100 @@ namespace Sapphire::Network::Packets::Server
...
+    uint32_t itemCatalogId;
+    uint32_t unknown1; // does some shit if nonzero
+    uint16_t requestId;
+    uint16_t quantity; // high/low u8s read separately?
+    uint32_t unknown3;
+  };
+
+  struct FFXIVIpcMarketBoardItemListing : FFXIVIpcBasePacket< MarketBoardItemListing >
+  {
+    struct itemListing // 152 bytes each
```
NotAdam:
```suggestion
    struct ItemListing // 152 bytes each
```

src/common/Network/PacketDef/Zone/ServerZoneDef.h
karashiiro: It always contains the same data looped 1-3 times, can't figure out what it is though

src/common/Network/PacketDef/Ipcs.h
karashiiro: Confirmed ItemListingCount, ListingHistory is not 13D. Filtering the packets is giving me a garbage opcode for ListingHistory, still need to fix that bug on my end

src/common/Network/PacketDef/Zone/ServerZoneDef.h
karashiiro: Just double-checked, that's always empty. padding6 is also always empty. I don't know what content IDs are supposed to look like, but parsing unknown8 through unknown10 as a uint64 gives numbers like `8444249366530776407` back, so maybe that's it?

NotAdam: it'll be requested via this actrl https://github.com/SapphireServer/Sapphire/blob/develop/src/common/Network/CommonActorControl.h#L322 and then you'll get this packet back https://github.com/SapphireServer/Sapphire/blob/develop/src/common/Network/PacketDef/Zone/ServerZoneDef.h#L1124-L1128

src/common/Network/PacketDef/Ipcs.h
NotAdam: huh, surprising that it's moved, wonder what's there now

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -283,6 +283,105 @@ namespace Sapphire::Network::Packets::Server
...
+      uint8_t padding5;
+      uint8_t retainerCity;
+      /**
+       * 0x01 Limsa Lominsa
+       * 0x02 Gridania
+       * 0x03 Ul'dah
+       * 0x04 Ishgard
+       * 0x07 Kugane
+       * 0x0A Crystarium
+       */
```
NotAdam: this should be above retainerCity, not below it

karashiiro: got it, i'll move the other one too

src/common/Network/PacketDef/Zone/ServerZoneDef.h
karashiiro: I don't seem to be getting back any packets with item creator names, which is strange, but there are definitely no outbound packets with that uint64, which rules that range out if nothing else

LimiQS:
[Here](https://github.com/menphina/Thaliak.Network/blob/master/Thaliak.Network/Messages/NetworkMarketListing.cs) is my structure definition on this message. Which might be useful.

karashiiro:
There's quite a bit there that this PR doesn't have, mind PRing [my fork](https://github.com/karashiiro/Sapphire) with that so it can get committed?

karashiiro:
Merged, thanks 👍

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -283,6 +283,105 @@ namespace Sapphire::Network::Packets::Server
...
+  };
+
+  struct FFXIVIpcMarketBoardItemListing : FFXIVIpcBasePacket< MarketBoardItemListing >
+  {
+    struct ItemListing // 152 bytes each
+    {
+      uint64_t listingId;
+      uint64_t retainerId;
+      uint64_t retainerOwnerId;
+      uint64_t artisanId;
```
NotAdam: while correct, the actual name for this is `contentId`

NotAdam: I wonder if SE uses the same ids for retainers too, can probably leave this as is actually

LimiQS: Ids of `retainerId` is exactly the same as the one in [NetworkRetainerSummary](https://github.com/menphina/Thaliak.Network/blob/master/Thaliak.Network/Messages/NetworkRetainerSummary.cs) when the item is listed by the retainer.

LimiQS: I don't have much study on the intrinsic connection of these ids or how SE stores these data. So `contentId` should be a better name.

NotAdam: makes sense - SE uses u64s to identify players globally (as in, cross datacenter) and it makes sense that retainers would be using the same IDs as well

PR #584, opened 2019-10-08
Adding ActorControl 20F packet def

Supamiu:
This packet is sent by server for desynth result, using 3 different forms from what I've seen so far (based on param1 value)

### DIFF
--- a/src/common/Network/PacketDef/Zone/ServerZoneDef.h
+++ b/src/common/Network/PacketDef/Zone/ServerZoneDef.h
@@ -519,7 +519,7 @@ namespace Sapphire::Network::Packets::Server
   {
     /* 0000 */ uint16_t category;
     /* 0002 */ uint16_t padding;
-    /* 0004 */ uint32_t resultType; // 4321 => item used, 4922 => item obtained, 4925 => exp obtained
+    /* 0004 */ Common::DesynthPacketType resultType;
     /* 0008 */ uint32_t param2;
     /* 000C */ uint32_t itemId;
     /* 0010 */ uint32_t param4;
--- a/src/common/Network/PacketDef/Zone/ServerZoneDef.h
+++ b/src/common/Network/PacketDef/Zone/ServerZoneDef.h
@@ -515,11 +515,11 @@ namespace Sapphire::Network::Packets::Server
   * Structural representation of the packet sent by the server
   * to update certain player details / status
   */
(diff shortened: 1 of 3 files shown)


src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -511,6 +511,23 @@ namespace Sapphire::Network::Packets::Server
     /* 0018 */ uint64_t targetId;
   };
 
+  /**
+  * Structural representation of the packet sent by the server
+  * to update certain player details / status
+  */
+  struct FFXIVIpcActorControl20F : FFXIVIpcBasePacket< ActorControl20F >
```
SapphireMordred: I don't think ActorControl is a good name for this.
ActorControls are very generic packets, this is way more specific.

NotAdam: the opcode is missing in ipcs.h too

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -511,6 +511,23 @@ namespace Sapphire::Network::Packets::Server
...
+  {
+    /* 0000 */ uint16_t category;
+    /* 0002 */ uint16_t padding;
+    /* 0004 */ Common::DesynthPacketType resultType;
```
NotAdam: is this actually logmessage or something?

Supamiu: It seems like it's used for both log message and the dialog box that appears to show the item you just desynthed and the result items. I am going to check later today if it's used for reduction result box too, as there's some unused fields that could just be related to reduction (collectability for instance)

src/common/Network/PacketDef/Zone/ServerZoneDef.h
Supamiu: Oh I thought I had to name it like this because xivmon was naming it `ActorCtrl<Unk:020F>`, something like that. I'll rename it properly and add the opcode to Ipcs.h then 👍

NotAdam: that's an actual actorcontrol then which are defined in this file https://github.com/SapphireServer/Sapphire/blob/develop/src/common/Network/CommonActorControl.h

Supamiu:
I'm closing this PR in order to create a new, cleaner one, using the correct implementation for this ActorControl packet.

PR #588, opened 2019-10-15
VS2019 support, end of VS2017 support

NotAdam:
gcc 7 doesn't support std::filesystem so there's a required update to 8 for it to build properly

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -20,8 +20,8 @@ before_build:
   - git submodule update --init
   - mkdir build
   - cd build
-  - cmake .. -G "Visual Studio 16 2019" -A Win64
-  - cmake --build . --target ALL_BUILD --config Release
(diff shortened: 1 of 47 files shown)


Lucyfurnice:
Also worth noting the compiler says _SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING define will silence the deprecation warning, since std::filesystem will replace experimental.
boostorg/boost filesystem might replace it for those who don't want to upgrade.

AzukaChan:
Any idea on when Sapphire will start VS2019 support? I'm aware we can use some of the VS2017 files on VS2019 (or just stick with VS2017) but it would be nice to be able to build Sapphire with the latest Visual Studio.

collett8192:
You can add one line in cmake [like this](https://github.com/collett8192/Sapphire/blob/develop/cmake/compiler.cmake#L9) to make vs2019 happy.

PR #607, opened 2019-12-25
Develop

Kirrilius:
fix warnings:
- using static casts
- comparing floating point with "==" or "!=" is not correct (Chara.cpp)

### DIFF
--- a/src/world/Action/Action.cpp
+++ b/src/world/Action/Action.cpp
@@ -45,14 +45,14 @@ Action::Action::Action( Entity::CharaPtr caster, uint32_t actionId, uint16_t seq
 
 Action::Action::Action( Entity::CharaPtr caster, uint32_t actionId, uint16_t sequence,
                         Data::ActionPtr actionData, FrameworkPtr fw ) :
-  m_pSource( std::move( caster ) ),
-  m_pFw( std::move( fw ) ),
-  m_actionData( std::move( actionData ) ),
   m_id( actionId ),
-  m_targetId( 0 ),
+  m_sequence( sequence ),
   m_startTime( 0 ),
+  m_pSource( std::move( caster ) ),
+  m_targetId( 0 ),
   m_interruptType( Common::ActionInterruptType::None ),
-  m_sequence( sequence )
+  m_pFw( std::move( fw ) ),
+  m_actionData( std::move( actionData ) )
 {
 }
 
@@ -101,10 +101,11 @@ bool Action::Action::init()
       case Common::ClassJob::Bard:
       case Common::ClassJob::Archer:
         m_range = 25;
-
+            break;
         // anything that isnt ranged
       default:
         m_range = 3;
+            break;
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/world/Actor/Chara.h
```diff
@@ -183,7 +183,7 @@ namespace Sapphire::Entity
 
     std::string getName() const;
 
-    bool face( const Common::FFXIVARR_POSITION3& p );
+    bool face( const Common::FFXIVARR_POSITION3& p , float epsilon = 0.005f);
```
NotAdam: don't do it this way, if you want to check the epsilon use `std::numeric_limits< float >::epsilon` instead of hardcoding it in the prototype

src/world/Manager/ActionMgr.cpp
```diff
@@ -78,7 +78,7 @@ void World::Manager::ActionMgr::handleItemAction( Sapphire::Entity::Player& play
 
 void World::Manager::ActionMgr::bootstrapAction( Entity::Player& player,
                                                  Action::ActionPtr currentAction,
-                                                 Data::Action& actionData )
+                                                 Data::Action& /*actionData*/ )
```
NotAdam: why?

src/world/Territory/Cell.cpp
```diff
@@ -11,10 +11,10 @@
 Sapphire::Cell::Cell() :
   m_bActive( false ),
   m_bLoaded( false ),
-  m_playerCount( 0 ),
-  m_bUnloadPending( false )
+  m_bUnloadPending( false ),
+  m_playerCount( 0 )
 {
-  m_bForcedActive = false;
+    m_bForcedActive = false;
```
NotAdam: This goes for pretty much everywhere else too, but please configure whatever you're using to respect `.editorconfig` or see the style guide

src/world/Action/Action.cpp
```diff
@@ -565,7 +566,7 @@ bool Action::Action::primaryCostCheck( bool subtractCosts )
   }
 }
 
-bool Action::Action::secondaryCostCheck( bool subtractCosts )
+bool Action::Action::secondaryCostCheck( bool /*subtractCosts*/ )
```
NotAdam: why?

src/world/Network/Handlers/ActionHandler.cpp
```diff
@@ -38,7 +38,7 @@ void Sapphire::Network::GameConnection::actionHandler( FrameworkPtr pFw,
   {
     default:
     {
-      player.sendDebug( "Skill type {0} not supported. Defaulting to normal action", type );
+      player.sendDebug( "Skill type {0} not supported. Defaulting to normal action", type ); break;
```
NotAdam: Why is there a break here? the message just before it indicates that this is intended behaviour

Kirrilius:
Sorry, it's my first pull request, i will do another pull request 
taking into account your comments

PR #620, opened 2020-01-05
Action system updates.

collett8192:
Combo system.
Aoe filter fixes, use AoeEffect packet whenever possible.
Basic self healing and mp restoring.
Enemy aggro on hostile actions.
Casting queue.
Interrupt cast if target died.
More opcode updates.


src/world/Action/Action.cpp
```diff
@@ -657,11 +747,24 @@ void Action::Action::addDefaultActorFilters()
 bool Action::Action::preFilterActor( Sapphire::Entity::Actor& actor ) const
 {
   auto kind = actor.getObjKind();
+  auto chara = actor.getAsChara();
 
   // todo: are there any server side eobjs that players can hit?
   if( kind != ObjKind::BattleNpc && kind != ObjKind::Player )
     return false;
 
+  if( m_lutEntry.potency > 0 && chara == m_pSource )
```
NotAdam: not sure if I'm a fan of comparing the ptrs, but I guess it'd work fine

src/world/Action/EffectBuilder.cpp
```diff
@@ -31,62 +31,225 @@ uint64_t EffectBuilder::getResultDelayMs()
...
+    EffectHeader* pHeader;
+    Common::EffectEntry* pEntry;
+    uint64_t* pEffectTargetId;
+    uint16_t* pFlag;
+    switch( packetSize )
+    {
+      case 8:
+      {
+        auto p = makeZonePacket< Server::FFXIVIpcAoeEffect8 >( m_sourceChara->getId() );
+        pHeader = ( EffectHeader* )( &( p->data() ) );
```
NotAdam: don't use c-style casts, keep it consistent with the rest of the codebase

src/world/Action/EffectBuilder.cpp
```diff
@@ -31,62 +31,224 @@ uint64_t EffectBuilder::getResultDelayMs()
...
+    pFlag[2] = 0x7FFF;
+
+    return effectPacket;
+  }
+  else
   {
-    auto result = it->second;
-    Logger::debug( " - id: {}", result->getTarget()->getId() );
+    auto resultList = m_resolvedEffects.begin()->second;
+    assert( resultList->size() > 0 );
```
NotAdam: you can just check if its `empty()` instead of `size() > 0`

src/world/Action/EffectResult.cpp
```diff
@@ -47,14 +48,36 @@ void EffectResult::damage( uint32_t amount, Common::ActionHitSeverityType severi
   m_type = Common::ActionEffectType::Damage;
 }
 
-void EffectResult::heal( uint32_t amount, Sapphire::Common::ActionHitSeverityType severity )
+void EffectResult::heal( uint32_t amount, Sapphire::Common::ActionHitSeverityType severity, bool isSelfHeal )
 {
   m_severity = severity;
   m_value = amount;
+  m_flag = isSelfHeal ? 0x80 : 0; // flag == 0x80 displays healing text at source actor
```
NotAdam: probably want to typedef these as there'd be other flags that are of interest to us too

    if( isSelfHeal )
      m_flag |= ShowFlyOverOnSource

makes a bit more sense too, given how it's also used elsewhere

src/world/Action/EffectResult.cpp
```diff
@@ -47,14 +48,36 @@ void EffectResult::damage( uint32_t amount, Common::ActionHitSeverityType severi
...
+
+void EffectResult::startCombo( uint16_t actionId )
+{
+  m_value = actionId;
+  m_flag = 0x80;
+
+  m_type = Common::ActionEffectType::StartActionCombo;
+}
+
+void EffectResult::comboVisualEffect()
```
NotAdam: is startCombo and comboVisualEffect mutually exclusive or something? maybe these names are a bit vague or I'm misunderstanding, but shouldn't there always be a visual effect? iirc when sending combos with effect you always had to send the successful combo action id

src/world/Actor/Player.cpp
```diff
@@ -2131,3 +2132,43 @@ Sapphire::Common::ActiveLand Sapphire::Entity::Player::getActiveLand() const
...
 }
+
+bool Sapphire::Entity::Player::hasQueuedAction() const
+{
+  return m_pQueuedAction != nullptr;
+}
+
+void Sapphire::Entity::Player::setQueuedAction( Sapphire::World::Action::ActionPtr pAction )
+{
+  m_pQueuedAction = nullptr; // overwrite whatever is already there
```
NotAdam: bit redundant cause you just nuke whatever was there anyway in the next line

src/world/Network/PacketWrappers/EffectPacket.h
```diff
@@ -25,13 +25,14 @@ namespace Sapphire::Network::Packets::Server
       m_data.effectTargetId = targetId;
 
       m_data.effectDisplayType = Common::ActionEffectDisplayType::ShowActionName;
+
+      std::memset( m_data.effects, 0, sizeof(Common::EffectEntry) * 8 );
```
NotAdam: code style, needs spacing in sizeof

src/world/Action/EffectResult.cpp
collett8192: StartActionCombo makes client flashes the next combo skill icon and ComboVisualEffect makes client uses a different skill effect (animation?) than using it uncomboed.

NotAdam: ah okay, so there'd be instances where you'd only have the visual effect, sounds good

src/world/Action/EffectResult.cpp
collett8192: I guess we don't want two healing functions so I merged them and let the caller decide who to heal and where to show flying text, same applied to mp restore and damage.

src/world/Action/EffectBuilder.cpp
```diff
@@ -31,62 +31,229 @@ uint64_t EffectBuilder::getResultDelayMs()
   return Common::Util::getTimeMs() + 850;
 }
 
-EffectResultPtr EffectBuilder::getResult( Entity::CharaPtr& chara )
+std::shared_ptr< std::vector< EffectResultPtr > > EffectBuilder::getResultList( Entity::CharaPtr& chara )
```
NotAdam: this doesn't really need to be a `shared_ptr`, should just use a `unique_ptr`

src/world/Action/EffectBuilder.cpp
```diff
@@ -31,62 +31,229 @@ uint64_t EffectBuilder::getResultDelayMs()
...
 
-  for( auto it = m_resolvedEffects.begin(); it != m_resolvedEffects.end(); )
+  auto globalSequence = m_sourceChara->getCurrentTerritory()->getNextEffectSequence();
+
+  do // we want to send at least one packet even nothing is hit so other players can see
   {
-    auto result = it->second;
-    Logger::debug( " - id: {}", result->getTarget()->getId() );
+    auto packet = buildNextEffectPacket( globalSequence );
+    m_sourceChara->sendToInRangeSet( packet, true );
```
NotAdam: the only problem with this is that `sourceSequence` can only be set for the caster, so you'd need to copy the packet here and zero that field when sending it to everyone else

src/world/Action/EffectBuilder.cpp
NotAdam: or just return a reference to the vector, bit nicer

src/world/Action/EffectBuilder.cpp
collett8192: that's not the case. sourceSequence is definitely counting up in other players' Effect as well. I can send a capture of me watching some random person hitting a fate monster if you want to see

NotAdam: hmm, alright, seem to recall that being the case previously but that was also more than a year ago when I actively looked into that packet so it could have changed or I'm remembering that wrong too

src/world/Action/EffectBuilder.cpp
collett8192: or not returning anything at all :)
see the latest update

NotAdam: lol, works for me, looks good

src/world/Action/EffectBuilder.cpp
collett8192: I did some tests there if you sent a sourceSequence = 0, or anything less than previous sent value for the same actor, the effect will not delay correctly, it will show effect immediately rather than sync with "the actor's weapon hitting the target". Maybe the client assumes it just lag and the effect is something already happened.

NotAdam: yeah, the sequence is used to basically sync up the animation on the caster, for example:

1. cast action a with seq = 2
2. cast succeeds, client automagically starts anim
3. client receives effect packet with caster seq = 3
4. client invalidates and starts playing new anim for seq 3 because to the client, it's not the same cast and effect packets take precedence on what is shown and will cancel out effects/anims if it is set in an effect packet

alternatively:

1. cast action a with seq = 2
2. cast succeeds, client automagically starts anim
3. client receives effect packet with caster seq = 2
4. seq matches most recent cast seq, current anim continues to play out with existing timing as its the same cast and nothing changes other than damage dealt being set to appear

PR #623, opened 2020-01-06
Allow action scripts to add effect.

collett8192:
and add the "+sprint" flying text effect when using sprint.

### DIFF
--- a/src/common/Common.h
+++ b/src/common/Common.h
@@ -622,6 +622,7 @@ namespace Sapphire::Common
     TpLoss = 12,
     TpGain = 13,
     GpGain = 14,
+    ApplyStatusEffect = 15,
     /*!
      * @brief Tells the client that it should show combo indicators on actions.
      *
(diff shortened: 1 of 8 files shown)


src/common/Common.h
```diff
@@ -668,13 +669,13 @@ namespace Sapphire::Common
...
     Common::ActionHitSeverityType hitSeverity;
-    uint8_t param;
+    uint8_t unk;
     /*!
      * @brief Shows an additional percentage in the battle log
      *
      * Has no effect on what is shown and stored in value
      */
-    int8_t bonusPercent;
+    uint8_t param; // this one is the real param
```
NotAdam: more of a note but depending on type, the next 2 (possibly 3 but never could confirm) bytes are basically parameters for the type handler

collett8192: I'm not exactly sure about this but this is where I found the "30" in the packet for sprint.

NotAdam: it's the effect param, in the case of sprint it's how fast you run

PR #624, opened 2020-01-06
Play the coin sound effect when buying items from shop.

collett8192:
When selling the sound effect is also played (and shop switch to buy back tab as well) but the selling logic code is not there so...

### DIFF
--- a/src/world/Actor/PlayerEvent.cpp
+++ b/src/world/Actor/PlayerEvent.cpp
@@ -145,8 +145,22 @@ void Sapphire::Entity::Player::playGilShop( uint32_t eventId, uint32_t flags, ui
   openGilShopPacket->data().eventId = eventId;
   openGilShopPacket->data().sceneFlags = flags;
   openGilShopPacket->data().actorId = getId();
-  openGilShopPacket->data().params[ 0 ] = 1;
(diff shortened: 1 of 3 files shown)


src/world/Actor/PlayerEvent.cpp
```diff
@@ -145,6 +145,8 @@ void Sapphire::Entity::Player::playGilShop( uint32_t eventId, uint32_t flags,
   openGilShopPacket->data().eventId = eventId;
   openGilShopPacket->data().sceneFlags = flags;
   openGilShopPacket->data().actorId = getId();
+  openGilShopPacket->data().params[ 0 ] = 1;
```
NotAdam: did I get the packet wrong? there's a field before this for the arg count

collett8192: this one is tricky. we only need the next param but this one cannot be 0. on retail it's 0x721A64A2 not looking like a count at all, and the count field we have is always 0

collett8192: ok the packet is wrong, see the latest update. I wonder why my wrong packet was working lol

NotAdam: yeah okay, maybe there's something else going on there, will need to double check the packet but this is fine for now

PR #625, opened 2020-01-07
Critical and direct hit.

collett8192:


src/common/Common.h
```diff
@@ -669,7 +669,7 @@ namespace Sapphire::Common
   {
     Common::ActionEffectType effectType;
     Common::ActionHitSeverityType hitSeverity;
-    uint8_t unk;
+    Common::ActionHitSeverityType healSeverity;
```
NotAdam: this isn't right in general, `hitSeverity`, `healSeverity` and `param` are basically like param0/1/2 and depends on what `effectType` is set to

collett8192: yea I'll rename them

PR #626, opened 2020-01-10
Generic action handler update. (please read description before looking into code)

collett8192:
The idea of this PR is from github.com/ravahn, he posted a message in discord pointing to some data he's being maintaining throughout his development of ACT FFXIV Plugin.

I built a new action parser https://github.com/collett8192/SapphireActionParseV2, extracts data to be used within Sapphire.

This means a big expansion to ActionLut, now containing status effect data for actions.

ActionLutData.cpp is still generated and kept untouched, but a few obvious error/missing data are found when debugging the parser and tweaked directly in parser code.

With this new data most statuses apply correctly, with the exception of non-100% proc, which in most case still need a check for unlock and will be scripted anyway.

Simple effects like dots, hots, damage buffs, defense buffs, heal buffs, crit or dh bonuses are working.

A few error in code are found and fixed during the implementation of this update.

With all being said, I'm totally good if Sapphire wants a different way to handle statuses, in which case I'll happily delete this and work with whatever the new implementation is.


src/world/Action/Action.cpp
```diff
@@ -809,11 +788,40 @@ Sapphire::Entity::CharaPtr Action::Action::getHitChara()
...
+  return isAttackTypeMagical( m_actionData->attackType );
+}
+
+bool Action::Action::isAttackTypePhysical( int8_t attackType )
+{
+  return attackType == -1 ||
+         attackType == 1 ||
+         attackType == 2 ||
+         attackType == 3 ||
+         attackType == 4;
```
Regiuz: Might wanna enum that

collett8192: exd structs should be generated however didn't find this one not sure if we should manually create enum or fix the generator

Regiuz: Ah, said struct only needs to be added to the enum generator then if it is present as exd.

Regiuz: https://github.com/SapphireServer/Sapphire/blob/master/src/tools/exd_common_gen/main.cpp#L117

collett8192: Fixed but the exd data maybe a little outdated and contains values that do not exist anymore. And it is in Japanese only so I had to manually edit it anyway.

NotAdam: I wouldn't put it inside commongen if you need to change it each time its regenerated, that defeats the purpose of it. just put it in common and remove it from the generator otherwise it'll just break

NotAdam:
I have a couple thoughts, mainly though I don't particularly like the idea of having to use data embedded as resources in another project because that's just then a lot of extra work if you want to use it as a data source. It'd be less of an issue if the ACT plugin was open source and you could just submodule it and away you go, but you can't so there's a bit of overhead there.

I don't think it's reasonable to have to download a new version of the plugin, extract and deflate the resource and then grab the xml (and other files too seeing as your tool uses them) files to create usable data. I think it's reasonable to use it as initial seed data, but I don't particularly like the structure of the files nor how they seem to be set out, so I wouldn't want to use them for anything more than that.

What I think there's merit in doing is providing this data in a more usable fashion and then making it easier for community contributions, which would not only benefit us but would likely benefit Ravahn too and would be a bit of a load off him. This would mean not using the lut in its current form, because for something like this I don't think it's a good idea for it to work in such a fashion given how the scope has increased since its initial conception as a stop gap solution, not something to provide functionality for a vast majority of actions.

I have a couple ideas I want to play with first before making an actual decision, but yeah, I think you're on the right track but I don't think this is the best way of achieving it.

collett8192:
I don't know the reason why he can't opensource his plugins, if he don't want to he wouldn't tell us about this in the first place. The extract & deflate can be automated and we can still submodule his repo to some degree, I just didn't bother implementing the extraction part because I knew you are not going to like it :)

Having a public repo that stores all these mappings then can be used by all ffxiv related project is a nice idea of course, however I don't think this project is active enough for something like that to happen any time soon and also be able to keep up to date, unless we just start with Ravahn's data.

His data contains the most info of what we can find now, at least it makes the game feel a lot better and could potentially attract more people into the project. We may want a better data source and parser that's for sure, well neither the data nor the parser is in this PR and nothing can prevent me from "handwriting" the lut table :)

src/world/Action/Action.cpp
```diff
@@ -809,11 +819,79 @@ Sapphire::Entity::CharaPtr Action::Action::getHitChara()
...
+
+void Action::Action::setPrimaryCost( Common::ActionPrimaryCostType type, uint16_t cost )
+{
+  m_primaryCostType = type;
+  m_primaryCost = cost;
+}
+
+Sapphire::StatusEffect::StatusEffectPtr Action::Action::createStatusEffect( uint32_t id, Entity::CharaPtr sourceActor, Entity::CharaPtr targetActor, uint32_t duration, uint32_t tickRate )
+{
+  // workaround to framework access issue in action script
```
NotAdam: or just do it properly?

src/world/Action/ActionLut.h
```diff
@@ -7,24 +7,43 @@ namespace Sapphire::World::Action
...
+    uint16_t targetStatus;
+    uint32_t targetStatusDuration;
+    uint16_t targetStatusParam;
+    uint16_t gainMPPercentage;
+  };
+
+  struct StatusEffectEntry
+  {
+    uint32_t effectType;
+    int32_t effectValue1;
```
NotAdam: what do these even mean? calling them all value is insanely obscure - if you're storing various kinds of data in there that have different meanings it makes much more sense to use a union

src/world/Action/ActionLut.h
```diff
@@ -7,24 +7,43 @@ namespace Sapphire::World::Action
...
-    uint16_t curePotency;
-    uint16_t restoreMPPercentage;
+    uint16_t damagePotency;
+    uint16_t damageComboPotency;
+    uint16_t damageDirectionalPotency;
+    uint16_t healPotency;
+    uint16_t selfHealPotency;
+    uint16_t selfStatus;
+    uint32_t selfStatusDuration;
+    uint16_t selfStatusParam;
```
NotAdam: this status stuff is really not great - see review comment

src/world/Action/EffectBuilder.cpp
```diff
@@ -49,65 +42,79 @@ void EffectBuilder::moveToResultList( Entity::CharaPtr& chara, EffectResultPtr r
   it->second->push_back( std::move( result ) );
 }
 
-void EffectBuilder::heal( Entity::CharaPtr& effectTarget, Entity::CharaPtr& healingTarget, uint32_t amount, Common::ActionHitSeverityType severity, Common::ActionEffectResultFlag flag )
+void EffectBuilder::heal( Entity::CharaPtr& effectTarget, Entity::CharaPtr& healingTarget, uint32_t amount, Common::ActionHitSeverityType severity, Common::ActionEffectResultFlag flag, uint64_t resultDelayMs )
 {
-  EffectResultPtr nextResult = make_EffectResult( healingTarget, getResultDelayMs() );
+  EffectResultPtr nextResult = make_EffectResult( healingTarget, nullptr, Common::Util::getTimeMs() + resultDelayMs );
```
NotAdam: why is this delay calculated manually each time?

src/world/Action/EffectBuilder.cpp
```diff
@@ -49,65 +42,79 @@ void EffectBuilder::moveToResultList( Entity::CharaPtr& chara, EffectResultPtr r
...
-  EffectResultPtr nextResult = make_EffectResult( damagingTarget, getResultDelayMs() );
+  EffectResultPtr nextResult = make_EffectResult( damagingTarget, nullptr, Common::Util::getTimeMs() + resultDelayMs );
   nextResult->damage( amount, severity, flag );
   moveToResultList( effectTarget, nextResult );
 }
 
 void EffectBuilder::startCombo( Entity::CharaPtr& target, uint16_t actionId )
 {
-  EffectResultPtr nextResult = make_EffectResult( target, 0 );
+  EffectResultPtr nextResult = make_EffectResult( target, nullptr, 0 );
```
NotAdam: if you're having to pass the same nullptr in every time, maybe consider another constructor which removes the need for said parameter - this is just more annoying to use

src/world/Action/EffectBuilder.cpp
```diff
@@ -49,65 +42,79 @@ void EffectBuilder::moveToResultList( Entity::CharaPtr& chara, EffectResultPtr r
...
 {
-  EffectResultPtr nextResult = make_EffectResult( target, 0 );
-  nextResult->applyStatusEffect( statusId, param );
+  EffectResultPtr nextResult = make_EffectResult( target, nullptr, 0 );
+  nextResult->statusNoEffect( statusId );
   moveToResultList( target, nextResult );
 }
 
-void EffectBuilder::buildAndSendPackets()
+void EffectBuilder::buildAndSendPackets( float animationLock )
```
NotAdam: again, why is this a parameter when it should just be a member var with a getter and setter?

src/world/Action/EffectBuilder.cpp
```diff
@@ -49,65 +42,79 @@ void EffectBuilder::moveToResultList( Entity::CharaPtr& chara, EffectResultPtr r
...
   {
-    auto packet = buildNextEffectPacket( globalSequence );
+    auto packet = buildNextEffectPacket( globalSequence, animationLock );
     m_sourceChara->sendToInRangeSet( packet, true );
   }
   while( !m_resolvedEffects.empty() );
 }
 
-std::shared_ptr< FFXIVPacketBase > EffectBuilder::buildNextEffectPacket( uint32_t globalSequence )
+std::shared_ptr< FFXIVPacketBase > EffectBuilder::buildNextEffectPacket( uint32_t globalSequence, float animationLock )
```
NotAdam: anim lock should be a member and used where applicable

src/world/Action/EffectBuilder.h
```diff
@@ -13,30 +13,30 @@ namespace Sapphire::World::Action
 
     void heal( Entity::CharaPtr& effectTarget, Entity::CharaPtr& healingTarget, uint32_t amount,
                Common::ActionHitSeverityType severity = Common::ActionHitSeverityType::NormalHeal,
-               Common::ActionEffectResultFlag flag = Common::ActionEffectResultFlag::None);
+               Common::ActionEffectResultFlag flag = Common::ActionEffectResultFlag::None, uint64_t resultDelayMs = 600 );
```
NotAdam: why is this hardcoded individually and differently in every function?

src/world/Actor/BNpc.cpp
```diff
@@ -679,34 +680,36 @@ void Sapphire::Entity::BNpc::setFlag( uint32_t flag )
...
 
-    auto effectPacket = std::make_shared< Server::EffectPacket >( getId(), pTarget->getId(), 7 );
-    effectPacket->setRotation( Util::floatToUInt16Rot( getRot() ) );
-    Common::EffectEntry effectEntry{};
-    effectEntry.value = static_cast< int16_t >( damage.first );
-    effectEntry.effectType = ActionEffectType::Damage;
-    effectEntry.param0 = static_cast< uint8_t >( damage.second );
-    effectEntry.param2 = 0x71;
-    effectPacket->addEffect( effectEntry );
+    auto exdData = m_pFw->get< Data::ExdDataGenerated >();
```
NotAdam: feels spaghetti, shouldn't be this kind of code here

src/world/Actor/Chara.cpp
```diff
@@ -663,27 +646,70 @@ void Sapphire::Entity::Chara::sendStatusEffectUpdate()
...
+    statusEffectList->data().effect[ slot ].sourceActorId = statusEffect->getSrcActorId();
     slot++;
   }
 
-  sendToInRangeSet( statusEffectList, isPlayer() );
+  if( totalShieldValue > 0 )
+  {
+    totalShieldValue /= getMaxHp();
+    totalShieldValue *= 100;
+    statusEffectList->data().shieldPercentage = totalShieldValue >= 255 ? 255 : static_cast< uint8_t >( totalShieldValue );
```
NotAdam: as above, just use std::min

src/world/Actor/Chara.cpp
```diff
@@ -551,28 +517,44 @@ void Sapphire::Entity::Chara::addStatusEffect( StatusEffect::StatusEffectPtr pEf
...
+      totalShieldValue += statusEffect->getEffectEntry().effectValue1;
+    }
+  }
+
+  if( totalShieldValue > 0 )
+  {
+    totalShieldValue /= getMaxHp();
+    totalShieldValue *= 100;
+    statusEffectAdd->data().shieldPercentage = totalShieldValue >= 255 ? 255 : static_cast< uint8_t >( totalShieldValue );
+  }
```
NotAdam: why is all this code duplicated again later in chara?

https://github.com/SapphireServer/Sapphire/pull/626/files#diff-0a3d3fe0bb35313cf639920ab3545a7cR689

NotAdam: also use std::min instead of this ternary crap

src/world/Actor/Chara.cpp
```diff
@@ -663,27 +646,70 @@ void Sapphire::Entity::Chara::sendStatusEffectUpdate()
...
+    totalShieldValue *= 100;
+    statusEffectList->data().shieldPercentage = totalShieldValue >= 255 ? 255 : static_cast< uint8_t >( totalShieldValue );
+  }
+
+  sendToInRangeSet( statusEffectList, true );
+}
+
+void Sapphire::Entity::Chara::sendEffectResultToUpdateShieldValue()
+{
+  auto pPacket = makeZonePacket< FFXIVIpcEffectResult >( getId() );
```
NotAdam: all this shouldn't be hardcoded in here, it's messy

src/world/Actor/Chara.cpp
```diff
@@ -907,6 +906,18 @@ uint32_t Sapphire::Entity::Chara::getStatValue( Sapphire::Common::BaseParam base
       break;
     }
 
+    case Common::BaseParam::Haste:
+    {
+      value = m_baseStats.haste;
+      for( auto const& statusIt : m_statusEffectMap )
+      {
+        auto effectEntry = statusIt.second->getEffectEntry();
+        if( static_cast< Common::StatusEffectType >( effectEntry.effectType ) == Common::StatusEffectType::Haste )
```
NotAdam: why isn't `effectType` a `Common::StatusEffectType`?

src/world/Actor/Player.cpp
```diff
@@ -1567,52 +1556,31 @@ uint32_t Sapphire::Entity::Player::getPersistentEmote() const
...
-    Common::EffectEntry entry{};
-    entry.value = damage.first;
-    entry.effectType = Common::ActionEffectType::Damage;
-    entry.param0 = static_cast< uint8_t >( damage.second );
-    entry.param2 = 0x72;
-
-    effectPacket->addEffect( entry );
-
-    sendToInRangeSet( effectPacket, true );
+    auto actionData = exdData->get< Data::Action >( 8 );
```
NotAdam: like bnpc, this is spaghetti and shouldn't be done inline

src/world/Network/Handlers/ClientTriggerHandler.cpp
```diff
@@ -134,7 +134,7 @@ void Sapphire::Network::GameConnection::clientTriggerHandler( FrameworkPtr pFw,
     case ClientTriggerType::RemoveStatusEffect: // Remove status (clicking it off)
     {
       // todo: check if status can be removed by client from exd
-      player.removeSingleStatusEffectById( static_cast< uint32_t >( param1 ) );
+      player.removeSingleStatusEffectById( static_cast< uint32_t >( param1 ), true, true );
```
NotAdam: either overload or default the values, it's pretty pointless to be setting them like this

src/world/StatusEffect/StatusEffect.cpp
```diff
@@ -85,33 +135,84 @@ uint16_t Sapphire::StatusEffect::StatusEffect::getParam() const
...
+        if( static_cast< Common::StatusEffectType >( effectEntry.effectType ) != Common::StatusEffectType::DamageMultiplier )
+          continue;
+        if( effectEntry.effectValue1 & m_effectEntry.effectValue1 )
+        {
+          damage *= 1.0f + ( effectEntry.effectValue2 / 100.0f );
+        }
+      }
+
+      m_value = Sapphire::Math::CalcStats::applyDamageReceiveMultiplier( *m_targetActor, damage,
+        m_effectEntry.effectValue1 == static_cast< int32_t >( Common::ActionTypeFilter::Physical ) ? Common::AttackType::Physical :
```
NotAdam: don't do this

src/world/StatusEffect/StatusEffect.cpp
```diff
@@ -85,33 +135,84 @@ uint16_t Sapphire::StatusEffect::StatusEffect::getParam() const
...
+      auto wepDmg = Sapphire::Math::CalcStats::getWeaponDamage( m_sourceActor );
+      auto heal = Sapphire::Math::CalcStats::calcHealBaseOnPotency( *m_sourceActor, m_effectEntry.effectValue2, wepDmg );
+
+      if( m_effectEntry.effectValue1 == 0 ) // this value is always 0 atm, if statement here just in case there is a hot that isn't a "cast"
+      {
+        for( auto const& entry : m_sourceActor->getStatusEffectMap() )
+        {
+          auto status = entry.second;
+          auto effectEntry = status->getEffectEntry();
+          if( static_cast< Common::StatusEffectType >( effectEntry.effectType ) != Common::StatusEffectType::HealCastMultiplier )
```
NotAdam: change `effectType` to be `Common::StatusEffectType`

src/world/StatusEffect/StatusEffect.cpp
```diff
@@ -85,33 +135,84 @@ uint16_t Sapphire::StatusEffect::StatusEffect::getParam() const
...
+        ( m_effectEntry.effectValue1 == static_cast< int32_t >( Common::ActionTypeFilter::Magical ) ? Common::AttackType::Magical : Common::AttackType::Unknown_0 ) );
+      m_cachedSourceCrit = Sapphire::Math::CalcStats::criticalHitProbability( *m_sourceActor, Common::CritDHBonusFilter::Damage );
+      m_cachedSourceCritBonus = Sapphire::Math::CalcStats::criticalHitBonus( *m_sourceActor );
+      break;
+    }
+
+    case Common::StatusEffectType::Hot:
+    {
+      auto wepDmg = Sapphire::Math::CalcStats::getWeaponDamage( m_sourceActor );
+      auto heal = Sapphire::Math::CalcStats::calcHealBaseOnPotency( *m_sourceActor, m_effectEntry.effectValue2, wepDmg );
```
NotAdam: why are these calculations duplicated even though they're used in the exact same way in both handlers?

src/world/StatusEffect/StatusEffect.cpp
```diff
@@ -85,33 +135,84 @@ uint16_t Sapphire::StatusEffect::StatusEffect::getParam() const
...
+    }
+  }
 }
 
 void Sapphire::StatusEffect::StatusEffect::removeStatus()
 {
   auto pScriptMgr = m_pFw->get< Scripting::ScriptMgr >();
   pScriptMgr->onStatusTimeOut( m_targetActor, m_id );
+
+  switch( static_cast< Common::StatusEffectType >( m_effectEntry.effectType ) )
```
NotAdam: fix type

src/world/Actor/Chara.cpp
```diff
@@ -551,28 +517,44 @@ void Sapphire::Entity::Chara::addStatusEffect( StatusEffect::StatusEffectPtr pEf
...
   auto effect = StatusEffect::make_StatusEffect( id, source.getAsChara(), getAsChara(), duration, 3000, m_pFw );
   effect->setParam( param );
+  effect->setLastTick( lastTickOverride == 0 ? Util::getTimeMs() : lastTickOverride );
   addStatusEffect( effect );
 }
 
-/*! \param StatusEffectPtr to be applied to the actor */
-void Sapphire::Entity::Chara::addStatusEffectByIdIfNotExist( uint32_t id, int32_t duration, Entity::Chara& source,
-                                                             uint16_t param )
+void Sapphire::Entity::Chara::addStatusEffectByIdIfNotExist( uint32_t id, int32_t duration, Entity::Chara& source, uint16_t param, uint64_t lastTickOverride )
```
NotAdam: why is the last tick overridden? if you're refreshing a status you should just update the timer on the existing status and resend that instead of constructing new objects everywhere

src/world/Actor/Chara.cpp
```diff
@@ -593,34 +575,36 @@ void Sapphire::Entity::Chara::statusEffectFreeSlot( uint8_t slotId )
...
     {
-      removeStatusEffect( effectIt.first );
+      removeStatusEffect( effectIt.first, sendActorControl, sendStatusList );
       break;
     }
   }
 }
 
-void Sapphire::Entity::Chara::removeStatusEffect( uint8_t effectSlotId )
+void Sapphire::Entity::Chara::removeStatusEffect( uint8_t effectSlotId, bool sendActorControl, bool sendStatusList )
```
NotAdam: either overload or default the params

src/world/Actor/Chara.cpp
```diff
@@ -593,34 +575,36 @@ void Sapphire::Entity::Chara::statusEffectFreeSlot( uint8_t slotId )
   m_statusEffectFreeSlotQueue.push( slotId );
 }
 
-void Sapphire::Entity::Chara::removeSingleStatusEffectById( uint32_t id )
+void Sapphire::Entity::Chara::removeSingleStatusEffectById( uint32_t id, bool sendActorControl, bool sendStatusList )
```
NotAdam: either overload or default the params

src/world/StatusEffect/StatusEffect.cpp
```diff
@@ -55,16 +68,53 @@ void Sapphire::StatusEffect::StatusEffect::registerTickEffect( uint8_t type, uin
...
+
+    case Common::StatusEffectType::Hot:
+    {
+      auto value = m_value;
+      if( m_cachedSourceCrit > Sapphire::Math::CalcStats::getRandomNumber0To99() )
+      {
+        value *= m_cachedSourceCritBonus;
+      }
+      value *= 1.0f + ( ( Sapphire::Math::CalcStats::getRandomNumber0To99() - 50.0f ) / 1000.0f );
+      m_currTickEffect = std::make_pair( 2, value );
```
NotAdam: use enums instead of literals, not sure what is with the obsession of using literals - it's really lazy and obscure

src/world/StatusEffect/StatusEffect.cpp
```diff
@@ -153,3 +254,75 @@ const std::string& Sapphire::StatusEffect::StatusEffect::getName() const
...
+  {
+    case Common::StatusEffectType::InstantCast:
+    {
+      if( !action->hasCastTime() )
+        return;
+      // value1: remaining uses
+      // value2-4: affected action ids, or all actions if value2 is 0
+      if( m_effectEntry.effectValue2 != 0 )
+      {
+        if( action->getId() != m_effectEntry.effectValue2 &&
```
NotAdam: this effectValue stuff is incredibly obscure, I don't like this as a solution but I don't think it'd be much better even if it were a union - see review message

src/scripts/action/sch/ActionAdloquium185.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  void onStart( Sapphire::World::Action::Action& action ) override
+  {
+    /*
+      don't run generic action handler for this action.
+      for simpler actions like cure 1 we only want to apply the freecure proc in script,
+      and let the generic handler do the heal so we don't have to copy heal code into scripts,
+      unless in cases like adlo, the healing result matters so we have to.
+    */
+    action.disableGenericHandler();
```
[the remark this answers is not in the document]
collett8192: it says clearly in the comment why we want to have both script and normal handling at the same time.

src/world/Actor/Chara.cpp
```diff
@@ -706,47 +732,20 @@ void Sapphire::Entity::Chara::updateStatusEffects()
...
-  if( thisTickHeal != 0 )
-  {
-    heal( thisTickDmg );
-    sendToInRangeSet( makeActorControl( getId(), HPFloatingText, 0,
-                                        static_cast< uint8_t >( ActionEffectType::Heal ), thisTickHeal ) );
   }
 }
 
-bool Sapphire::Entity::Chara::hasStatusEffect( uint32_t id )
+std::pair< uint8_t, Sapphire::StatusEffect::StatusEffectPtr > Sapphire::Entity::Chara::getStatusEffectById( uint32_t id )
```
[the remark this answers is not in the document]
collett8192: then you need to see code carefully

src/world/Actor/Chara.cpp
collett8192: it needs to recalculate damage/heal, reset uses, reapply shield etc, it's not a timer extend

src/world/Action/Action.cpp
collett8192: the framework in script is a raw pointer

src/world/Action/EffectBuilder.cpp
collett8192: it allows applying different delay for individual result, which is need for some scripts to use like summoner dots

src/world/Action/EffectBuilder.h
collett8192: just default testing value to use until real value is found.
However the 500 for status is very likely correct for single buff button.

src/world/Actor/Chara.cpp
collett8192: two different packets have this same shield field need to be sent, and not worth it to create a function for this since one of them had an existing loop can be used instead of looping it again just for this, and only clients care about this percentage anyway

src/world/StatusEffect/StatusEffect.cpp
collett8192: this is not added by me. I just return the value needed for it to work.

src/world/StatusEffect/StatusEffect.cpp
collett8192: I saw fields like actionCategory all using ints instead of enums in common, so I keep it consistent

src/world/StatusEffect/StatusEffect.cpp
collett8192: I agree with it however I don't think there's any better way than a union, and also easy to generate.

collett8192:
At this point it is just me having fun implement things to play around and I agree the code is not looking that good. Since @NotAdam started a new repo storing action data, I'll close this one.
I'm not 100% agree on all of the comments though. After the new data struct gets pushed into develop I'll port parts that didn't get a comment (hope it means good enough) and create new PRs.

src/world/Actor/Chara.cpp
NotAdam: ah, I missed that, reasonable but it would be better to have something to 'reset' a status rather than having to nuke it entirely and recreate it.

collett8192: I think just create a new one looks cleaner though

NotAdam: Maybe, but doing something like `status->refresh(...)` which in effect creates a 'new' status but still preserves needed data seems like a much cleaner approach to me

NotAdam: you don't have to alloc any more memory to do that either so it's a bit faster as well

collett8192: or is it? you are going to modify the target status map in the status itself, in most existing access to the map it's in a iteration and can't really modify the map anyway. there's a markToRemove flag I created so status itself don't have to care about those things.

NotAdam: you're not modifying the map if you refresh/modify the status directly, it's just a ptr whereas your add/remove does modify the map, though that's not really a concern of mine - it doesn't really matter.

src/scripts/action/sch/ActionAdloquium185.cpp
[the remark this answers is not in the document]
NotAdam: I _don't_ really like this because it seems more like a hack than anything else and I can see some merit in doing it this way but I'd be inclined to not do it this way because the split responsibility makes it less obvious what's going on. But I think if you did the inverse and explicitly enabled it from a script you'd have the same problem anyway so I guess it doesn't matter much

src/world/Actor/Chara.cpp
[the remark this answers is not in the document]
NotAdam: yeah after looking at it I don't think it's really good lol, ideally you should be able to remove statuses just with the object itself and not have to maintain slot numbers outside of networking code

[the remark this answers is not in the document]
NotAdam: just goes back to how the status system is really half baked at the moment and has really core issues that need to be resolved first

src/world/Actor/Chara.cpp
collett8192: oops I need to go back to school and learn English

src/world/Actor/Chara.cpp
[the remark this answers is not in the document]
collett8192: well I guess this is a non issue with a status->refresh addition

src/world/Actor/Chara.cpp
NotAdam: lmao you're fine, doesn't help that half the stuff I say makes zero sense. let me know if something isn't clear or whatever and I can try explaining it in a not retarded way

PR #628, opened 2020-01-11
Update EffectResult packet and send proper animation lock time.

collett8192:


src/world/Action/EffectBuilder.cpp
```diff
@@ -28,7 +28,7 @@ uint64_t EffectBuilder::getResultDelayMs()
 {
   // todo: actually figure this retarded shit out
 
-  return Common::Util::getTimeMs() + 850;
+  return Common::Util::getTimeMs() + 600;
```
NotAdam: More of an fyi than anything else, this isn't correct. Depends on the action itself so any value you put here for the time being will pretty much be wrong

src/world/Action/EffectBuilder.cpp
```diff
@@ -176,6 +176,7 @@ std::shared_ptr< FFXIVPacketBase > EffectBuilder::buildNextEffectPacket( uint32_
     pHeader->effectCount = static_cast< uint8_t >( remainingTargetCount > packetSize ? packetSize : remainingTargetCount );
     pHeader->sourceSequence = m_sequence;
     pHeader->globalSequence = globalSequence;
+    pHeader->animationLockTime = 0.6f;
```
NotAdam: This (and the other 2) are also wrong for the same reason. Regular attacks are 0.6s for the most part but a lot of others have different timings as well. Limit breaks are roughly 7.5s or so.

Believe they do this with an action category of sorts given how many shared values there are but I never really cared to confirm it

NotAdam: Also ninja attacks (or just mudras, been a while) are handled differently for these too.

collett8192: didn't get anything other than 0.6 when I was capturing. well lb will be one I guess.

src/world/Action/EffectBuilder.cpp
collett8192: this is just for the 0.6 animation lock time because you can't let the client be able to hit next button before previous effects execute otherwise it will show weird things like duplicated flying text. pretty sure caster sequence related.

NotAdam: This delay has nothing to do with animation lock, it's at what point the server should send a result which effects a targets attributes/statuses. if it doesn't match the animation time correctly, a target will lose hp before the flying text appears which looks wrong & not how retail works

NotAdam: Not saying you need to fix this, but just letting you know that this isn't correct nor how it works - it doesn't really matter what the value is at this point but just note that it depends on the action itself and is unrelated to anything else other than syncing up things applying when actions actually connect with a target

collett8192: I'm not saying it has anything to do with animation lock. It's because I set animation lock to 0.6 atm, this cannot be 0.85 anymore and have to be equal or less than 0.6, because If client hits the next button before receive previous effect it advances it's internal caster sequence then after that receives effect packet with smaller caster sequence.

collett8192: I guess I'm crazy. I should receive effect packet before that. Then I don't know what about the visual bug I got when playing with it.

collett8192: nvm that was when i'm playing with 200ms ping and no animation lock at all. so I should be animation lock > latency so sure nothing to do with this. At the same time 0.85 didn't feel right anyway.

src/world/Action/EffectBuilder.cpp
collett8192: Actually I got two non-0.6 results, see latest update.

src/world/Action/EffectBuilder.cpp
NotAdam: yeah this game quickly makes you question your sanity and everything else you knew. 

I haven't had a chance yet to take a good look at all of this, but basically what needs to happen/should happen is:
1. You calculate an effect and it's result(s).
2. You send the results in the initial packet `FFXIVIpcEffect` which plays the animation and sets up the battle text and so on, but it's purely visual.
3. You then cache those resultant effects and throw it somewhere with a x ms delay
4. You then apply damage/healing/status/etc to a target once the wait time has elapsed by applying it to the actor(s) in question and sending a `FFXIVIpcEffectResult`

Some notes though:
* To be retail accurate, the above (with 0 latency) should basically line up exactly with when the floating text pops up as it's what will actually show damage on the HP bar, or actually apply a status.
* With one exception, the summoner abilities Tri-disaster (and Bio/Miasma?) apply their statuses immediately now because SE finally realised that the delay made fester and bane 1000x more annoying to use, so there needs to be consideration there too. The potency damage on Tri-disaster is still delayed though iirc, so it's just annoying bullshit

850ms is correct for Glare and I believe Stone 1/2/3 as well, if you set it back to that and use those abilities you'll see it should sync up correctly, which is how it looks on retail with a good connection.

Ping doesn't affect anim lock, anim lock is basically a forced lockout on casting another action regardless of its cooldown state. It's not _really_ an animation lock per se, but I called it that a year ago when I was figuring out the packet and everyone else also calls it animation lock now, so I guess it should stay. Realistically it's more of an OGCD lock.

collett8192: Actually I'm leaning toward a constant value for this. Tried on retail, using stone the HP bar went down faster than text and using heavy swing text appeared faster than HP bar. I'm talking about very obvious desync not something latency should introduce.

collett8192: https://imgur.com/a/gd4ECMH
https://imgur.com/a/RRcmdf3
See it in action.

collett8192: https://imgur.com/a/mksxF9O
One more to show it's not a cast vs melee thing.
It seems they didn't even bother to sync it up. Longer animation damage first text second, shorter animation text first damage second.

collett8192:
I'm merging this into the action update pr because I need fix from this to implement more statuses :)

PR #629, opened 2020-01-19
enum/struct/packet updates, fix dot/hot tick, remove incorrect initial delay for player autoattacks.

collett8192:
Stuff picked up from closed PR.


NotAdam:
looks good, thanks for pulling all that out of the last PR

PR #637, opened 2020-02-18
5.2 packet update.

collett8192:
should be able to login and do basic stuff


NotAdam:
fix the build and I'll merge this in

collett8192:
Wait a moment before merging, there is another packet to fix.

collett8192:
Should be good now.

PR #643, opened 2020-02-25
Solve the mystery of second quest not showing in gridania or limsa.

collett8192:
I didn't bother trying to fix those offset values in the names of unknown fields. At this point this packet is just so broken. Well at least I tried to line up important stuff like levels, exp and player name.


NotAdam:
this isn't really a proper fix, but if it works it works. I'll take a proper look at fixing this packet soon:tm: probably

PR #651, opened 2020-03-14
implemented dyeable items

Arazati:
Functional, but doesn't remove any dyes on use, and doesn't line up 1:1 with retail packets I'm sure.
As a bonus, fixed up some of the Examine stuff so dyes show up properly when you examine a player.

### DIFF
--- a/src/world/Actor/Player.cpp
+++ b/src/world/Actor/Player.cpp
@@ -1952,7 +1952,7 @@ void Sapphire::Entity::Player::dyeItemFromDyeingInfo()
 
   // TODO: subtract/remove dye used
 
-  insertInventoryItem( static_cast < Sapphire::Common::InventoryType > ( itemToDyeContainer ), static_cast < uint16_t > ( itemToDyeSlot ), itemToDye );
+  insertInventoryItem( static_cast< Sapphire::Common::InventoryType >( itemToDyeContainer ), static_cast< uint16_t >( itemToDyeSlot ), itemToDye );
   writeItem( itemToDye );
 }
 
--- a/src/world/Actor/Player.cpp
+++ b/src/world/Actor/Player.cpp
@@ -1926,6 +1926,36 @@ uint8_t Sapphire::Entity::Player::getNextObjSpawnIndexForActorId( uint32_t actor
   return index;
 }
 
+void Sapphire::Entity::Player::setDyeingInfo( uint32_t itemToDyeContainer, uint32_t itemToDyeSlot, uint32_t dyeBagContainer, uint32_t dyeBagSlot)
+{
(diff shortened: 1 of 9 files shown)


src/world/Actor/Player.h
```diff
@@ -1114,6 +1117,14 @@ namespace Sapphire::Entity
 
     Common::PlayerTeleportQuery m_teleportQuery;
 
+    struct PlayerDyeingInfo
+    {
+      uint32_t itemToDyeContainer;
+      uint32_t itemToDyeSlot;
+      uint32_t dyeBagContainer;
+      uint32_t dyeBagSlot;
+    }; PlayerDyeingInfo m_dyeingInfo;
```
NotAdam: I would be inclined to put this struct elsewhere, though I'm not entirely sure on where this should go either. either way, the actual field should be on it's own line, not right after the semicolon

src/world/Actor/Player.h
```diff
@@ -536,6 +536,9 @@ namespace Sapphire::Entity
 
     void clearTeleportQuery();
 
+    void setDyeingInfo(uint32_t itemToDyeContainer, uint32_t itemToDyeSlot, uint32_t dyeBagContainer, uint32_t dyeBagSlot);
```
NotAdam: code style

src/world/Actor/Player.cpp
```diff
@@ -1926,6 +1926,36 @@ uint8_t Sapphire::Entity::Player::getNextObjSpawnIndexForActorId( uint32_t actor
...
+
+  if (!itemToDye || !dyeToUse) return;
+
+  uint32_t stainColorID = dyeToUse->getAdditionalData();
+  itemToDye->setStain(stainColorID);
+
+  //TODO: subtract/remove dye used
+
+  insertInventoryItem((Sapphire::Common::InventoryType)itemToDyeContainer, (uint16_t)itemToDyeSlot, itemToDye);
+  writeItem(itemToDye);
```
NotAdam: code style here too

src/world/Network/Handlers/ClientTriggerHandler.cpp
```diff
@@ -279,10 +279,7 @@ void Sapphire::Network::GameConnection::clientTriggerHandler( FrameworkPtr pFw,
     }
     case ClientTriggerType::DyeItem: // Dye item
     {
-      // param11 = item to dye container
-      // param12 = item to dye slot
-      // param2 = dye bag container
-      // param4 = dye bag slot
+      player.setDyeingInfo(param11, param12, param2, param4);
```
NotAdam: code style

PR #681, opened 2020-06-20
Class Quests ARC

DantestyleXD:
Hi everyone!

Finally I'm back to fix! 😃 

With this pull request Arc chain quest will be complete.

Just a couple of things to fix ( need someone better than me for this ):
 - Questbattles are broken: I can't init them that world.exe crash (Log trace below)
[questbattles.txt](https://github.com/SapphireServer/Sapphire/files/4807183/questbattles.txt)
Last try at commit 7ac677c70f8f6b6b8cdd7b3ee79dab43a32e4a94

 - EventActionStart on Eobj doesn't work properly: player's animation start correctly but world.exe crash...
I found a temporary solution with "EventStart": it lets going well with the quest but appear a "Event Cancelled" red warning everytime that call function.

- Need an external text with ClsArc005: do ingame //gm quest seq 65606 4 and talk to eobj for complete the task, go and try to talk with Buscaron;

### DIFF
--- a/src/scripts/quest/classquest/ARC/ClsArc006.cpp
+++ b/src/scripts/quest/classquest/ARC/ClsArc006.cpp
@@ -1,0 +1,265 @@
+#include <Actor/Player.h>
+#include "Manager/EventMgr.h"
+#include <ScriptObject.h>
+#include <Service.h>
+
+using namespace Sapphire;
+
+// Quest Script: ClsArc006_00076
+// Quest Name: The One That Got Away
+// Quest ID: 65612
+// Start NPC: 1000200
+// End NPC: 1000200
+
+class ClsArc006 : public Sapphire::ScriptAPI::EventScript
+{
+   private:
+      // Basic quest information 
+      // Quest vars / flags used
+      // GetQuestBitFlag8
+      // GetQuestUI8AL
+
+      enum Sequence : uint8_t
+      {
+         Seq0 = 0,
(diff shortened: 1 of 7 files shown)


src/scripts/quest/classquest/ARC/ClsArc006.cpp
```diff
@@ -0,0 +1,265 @@
...
+   void Scene00002( Entity::Player& player )
+   {
+      player.playScene( getId(), 2, HIDE_HOTBAR,
+                        [ & ]( Entity::Player& player, const Event::SceneResult& result )
+                        {
+                          if( result.param2 == 1 )
+                          {
+                              player.updateQuest( getId(), Seq2 );
+                              // QuestBattle doesn't work
+                              /*auto& pTeriMgr = Common::Service< Sapphire::World::Manager::TerritoryMgr >::ref();
```
collett8192, line 128: Getting services from script currently not working if built on windows.

DantestyleXD: I took it from ManFst005 since it worked.
Do I have to wait service fix or is there an alternative way?

collett8192: Code is fine, it's a compiler problem. If you want to test it on Windows then you have to get the service object from a different source (e.g. player) by adding a temp function and call ref() there then return the result to the script.

PR #687, opened 2020-07-02
Implement party just enough to gpose.

collett8192:
I don't think this should be merged because of the garbage party implementation.
Just leave here for packet information in case someone decides to write a proper PartyMgr.

Or gposers to merge by themselves :)


collett8192:
Since this is on 5.25 thus not working anymore. Packet structures are already in develop (without handlers).
A 5.35 party enabled gpose server can be found [here](https://github.com/collett8192/Sapphire/tree/develop_5.35u) :)

PR #691, opened 2020-08-15
Remove UseMooch since the detection for this is actually InventoryTransaction and the packet is SomeDirectorUnk4 or something

karashiiro:
Anyways, this one isn't supposed to be here.


karashiiro:
Tossing in SkillHandler = 0x033C from Winter while we're at it <:pepefeet:656180575262212151>

PR #705, opened 2021-01-09
Add support for residential ward house listing packets

p-doyle:
These are the packets sent when viewing a ward in the residential district menu.  Unfortunately I do not have a way to test the struct but it should build at least...  I have tested an equivalent struct in python and nodejs though.  Also I wasn't sure how to handle the houseSize byte because it's also needed for the housePrice uint32... 

This is my first attempt at this, please be gentle!

### DIFF
--- a/src/common/Network/PacketDef/Ipcs.h
+++ b/src/common/Network/PacketDef/Ipcs.h
@@ -255,6 +255,7 @@ namespace Sapphire::Network::Packets
 
     LandSetMap = 0x025B, // updated 4.5
 
+    ResidentialWardHouseListing = 0x0096,   // added 5.4
     //////////////////////////////////////////////////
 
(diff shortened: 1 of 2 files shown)


collett8192:
looks like [this one](https://github.com/SapphireServer/Sapphire/blob/develop/src/common/Network/PacketDef/Zone/ServerZoneDef.h#L2010)?

p-doyle:
Ugh sorry, I swear I looked for an existing one but not hard enough apparently...  

Is it worth updating the PR to just update the IPC code for HousingWardInfo?  Or just close the PR and slink away in shame?

collett8192:
Probably not updating it atm. Current code still targeting 5.35 hotfix and this packet is used in the code.

NotAdam:
> Is it worth updating the PR to just update the IPC code for HousingWardInfo?

yes, feel free

PR #706, opened 2021-01-23
5.35 hotfix opcodes

JeidoUran:
Not sure if this is really useful since this is for a now older version of the game, but it should at least help making housing usable on 5.35 (there is still a crash upon entering houses, but it seems unrelated to opcodes).


collett8192:
Does the server crashes or the client?

JeidoUran:
Server. Seems to be an issue with findOrCreateHousingInterior.

NotAdam:
probably exd related tbh

PR #715, opened 2021-07-28
Implement housing edit interior.

collett8192:
And some opcode updates thrown in as well.

### DIFF
--- a/src/world/Manager/HousingMgr.cpp
+++ b/src/world/Manager/HousingMgr.cpp
@@ -1663,7 +1663,7 @@ void Sapphire::World::Manager::HousingMgr::editAppearance( bool isInterior, Sapp
           if( oldItem )
           {
             housingContainer->removeItem( i );
-            invMgr.removeItemFromHousingContainer( land->getLandIdent(), housingContainer->getId(), i );
+            invMgr.removeItemFromHousingContainer( landIdent, housingContainer->getId(), i );
             player.addItem( oldItem, false, false, false );
           }
         }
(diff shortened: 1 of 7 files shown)


src/world/Manager/HousingMgr.h
```diff
@@ -181,7 +181,7 @@ namespace Sapphire::World::Manager
 
     bool hasPermission( Entity::Player& player, Land& land, uint32_t permission );
 
-    void editExterior( Sapphire::Entity::Player& player, uint16_t plot, std::vector< uint16_t > containerList, std::vector< uint8_t > slotList, uint8_t removeFlag );
+    void editAppearance( bool isInterior, Sapphire::Entity::Player& player, const Common::LandIdent& landIdent, std::vector< uint16_t > containerList, std::vector< uint8_t > slotList, uint8_t removeFlag );
```
NotAdam, line 184: there's no point of this being a reference, it's the size of a ptr and you don't change the original so it's just wasteful

src/world/Manager/HousingMgr.cpp
```diff
@@ -1632,7 +1632,7 @@ Sapphire::Inventory::HousingItemPtr Sapphire::World::Manager::HousingMgr::getHou
   return Inventory::make_HousingItem( tmpItem->getUId(), tmpItem->getId() );
 }
 
-void Sapphire::World::Manager::HousingMgr::editAppearance( bool isInterior, Sapphire::Entity::Player& player, const Common::LandIdent& landIdent, std::vector< uint16_t > containerList, std::vector< uint8_t> slotList, uint8_t removeFlag )
+void Sapphire::World::Manager::HousingMgr::editAppearance( bool isInterior, Sapphire::Entity::Player& player, Common::LandIdent landIdent, std::vector< uint16_t > containerList, std::vector< uint8_t> slotList, uint8_t removeFlag )
```
NotAdam, line 1635: it should still be const though...

collett8192: It shouldn't matter since it copies the value so the caller cannot be affected?

NotAdam: not relevant nor a good excuse, if you're not changing it it should be const. it provides compile time benefits too if you mark an parameteras const as it can perform optimisations that could otherwise be impossible/difficult to deduce with dfa, e.g reusing the value between calls/nested function calls because the compiler is guaranteed that its value couldn't have changed

PR #720, opened 2021-08-15
PublicContent support.

collett8192:


src/common/Network/PacketDef/Ipcs.h
```diff
@@ -194,14 +194,7 @@ namespace Sapphire::Network::Packets
     EventPlay128 = 0x026E, // updated 5.58
     EventPlay255 = 0x039E, // updated 5.58
 
-    EventYield = 0x0123, // updated 5.58
```
NotAdam, line 197: why is this removed and then added back later in a different spot?

src/world/Actor/Player.cpp
```diff
@@ -503,8 +511,18 @@ bool Sapphire::Entity::Player::exitInstance()
 {
   auto& teriMgr = Common::Service< TerritoryMgr >::ref();
 
-  auto pZone = getCurrentTerritory();
-  auto pInstance = pZone->getAsInstanceContent();
+  if( auto d = getCurrentTerritory()->getAsDirector() )
+    if( d->getContentFinderConditionId() > 0 )
```
NotAdam, line 515: dont put the assignment in the conditional, do it outside and then `d && d->blah`

NotAdam (changes requested):

src/common/Network/PacketDef/Ipcs.h
collett8192: This is renamed to EventContinue, a few lines below.

NotAdam:
```diff
diff --git a/src/common/Network/PacketDef/Ipcs.h b/src/common/Network/PacketDef/Ipcs.h
index c33cfad6e..7bce247ea 100644
--- a/src/common/Network/PacketDef/Ipcs.h
+++ b/src/common/Network/PacketDef/Ipcs.h
@@ -194,14 +194,7 @@ namespace Sapphire::Network::Packets
     EventPlay128 = 0x026E, // updated 5.58
     EventPlay255 = 0x039E, // updated 5.58
 
-    EventYield = 0x0123, // updated 5.58
-    //EventYield4 = 0x0000,
-    //EventYield8 = 0x0000,
-    //EventYield16 = 0x0000,
-    //EventYield32 = 0x0000,
-    //EventYield64 = 0x0000,
-    //EventYield128 = 0x0000,
-    //EventYield255 = 0x0000,
+    EventContinue = 0x0123, // updated 5.58
@@ -398,7 +393,8 @@ namespace Sapphire::Network::Packets
     EventYieldHandler = 0x03A2, // updated 5.58
     ReturnEventHandler = 0x0333, // updated 5.58
     TradeReturnEventHandler = 0x0179, // updated 5.58
-    TradeReturnEventHandler2 = 0x0169, // updated 5.55 hotfix
+    TradeReturnEventHandler2 = 0x02E1, // updated 5.58
+    EventYield16Handler = 0x03D7, // updated 5.58
```

you remove EventYield16 and add it back with a different name in a different place

eventcontinue != eventyield

collett8192: The EventYield16Handler is a newly found client packet, and (one of) its reply is renamed to EventContinue,

NotAdam: you literally removed `EventYield16` and then added it back in a different place as `EventYield16Handler`, it's not new, fix the opcodes if they're wrong instead of removing things entirely

collett8192: No, none of those opcodes are wrong, since the client "yields" from the event, I move the "EventYield" packet name to it, and the reply (0x0123) now needs a new name, that's EventContinue.

NotAdam: okay I'm retarded and can't read a diff

you're still removing `EventYield` entirely and replacing it with `EventContinue`, both of which exist

NotAdam: actually nevermind, I can't find where I had the capture for this in the past so just forget it

collett8192: If you are saying there is another (list of) packets should named EventYield on the SERVER side as well as EventContinue, then we don't know any opcode of them.

PR #721, opened 2021-08-30
Including support for basic map icons

Taezen:
I have wrote a map manager that manages the map icons for the big map.

It is supporting the following map icons for now:
  - Quest icons (only pre-accepted ones, because accepted ones are getting handled by the client)
  - Guildleve icons
  - Guildhest icons
  - Beginner arena icons
  - Triple Triad Master npc icon

For the quest icons I reversed engineered the quest availability check from the client. It doesn't check everything what the client does for now, but all that what is currently supported by the server.

### DIFF
--- a/src/common/Network/PacketDef/Zone/ServerZoneDef.h
+++ b/src/common/Network/PacketDef/Zone/ServerZoneDef.h
@@ -928,12 +928,12 @@ namespace Sapphire::Network::Packets::Server
 
     uint8_t unknown5;
     uint32_t unknown8;
-    uint16_t festivalId;
-    uint16_t additionalFestivalId;
     uint32_t unknown9;
     uint32_t unknown10;
+    uint32_t festivalId;
     uint32_t unknown11;
-    uint32_t unknown12[4];
+    uint32_t unknown12[3];
+    uint32_t additionalFestivalId;
     uint32_t unknown13[3];
     Common::FFXIVARR_POSITION3 pos;
     uint32_t unknown14[3];
@@ -2288,6 +2288,60 @@ namespace Sapphire::Network::Packets::Server
       uint8_t look[26];
       uint16_t padding3;
     } actors[2];
+  };
+
+  //For quests this is only used for pre-accepted ones. Accepted quests are getting handled by the client.
+  template< int ArgCount >
+  struct FFXIVIpcMapUpdateN
+  {
+    uint8_t entryCount;
+    uint8_t padding[ 3 ];
+    uint32_t iconIds[ ArgCount ];
+    uint32_t levelIds[ ArgCount ];
+    uint32_t eventIds[ ArgCount ]; // possible event ids for this: Quest, GuildLeveAssignment, GuildOrderGuide, TripleTriad, CustomTalk, PreHandler
+    uint8_t additionalData[ ArgCount ]; // use unknown
+  };
+
+  struct FFXIVIpcMapUpdate :
+    FFXIVIpcBasePacket< MapUpdate >,
+    FFXIVIpcMapUpdateN< 2 >
+  {
+  };
+
+  struct FFXIVIpcMapUpdate4 :
+    FFXIVIpcBasePacket< MapUpdate4 >,
+    FFXIVIpcMapUpdateN< 4 >
+  {
+  };
+
+  struct FFXIVIpcMapUpdate8 :
+    FFXIVIpcBasePacket< MapUpdate8 >,
+    FFXIVIpcMapUpdateN< 8 >
+  {
+  };
+
+  struct FFXIVIpcMapUpdate16 :
+    FFXIVIpcBasePacket< MapUpdate16 >,
+    FFXIVIpcMapUpdateN< 16 >
+  {
+  };
+
+  struct FFXIVIpcMapUpdate32 :
+    FFXIVIpcBasePacket< MapUpdate32 >,
+    FFXIVIpcMapUpdateN< 32 >
+  {
+  };
+
+  struct FFXIVIpcMapUpdate64 :
+    FFXIVIpcBasePacket< MapUpdate64 >,
+    FFXIVIpcMapUpdateN< 64 >
+  {
+  };
+
+  struct FFXIVIpcMapUpdate128 :
+    FFXIVIpcBasePacket< MapUpdate128 >,
+    FFXIVIpcMapUpdateN< 128 >
+  {
   };
 }
 
--- a/src/common/Network/PacketDef/Zone/ServerZoneDef.h
+++ b/src/common/Network/PacketDef/Zone/ServerZoneDef.h
@@ -931,7 +931,6 @@ namespace Sapphire::Network::Packets::Server
     uint32_t unknown9;
     uint32_t unknown10;
     uint32_t festivalId;
-    uint32_t unknown11;
     uint32_t unknown12[3];
     uint32_t additionalFestivalId;
     uint32_t unknown13[3];
--- a/src/world/Event/EventHandler.h
+++ b/src/world/Event/EventHandler.h
@@ -85,6 +85,13 @@ namespace Sapphire::Event
       QuestBattleDirector = 0x8006,
     };
 
+    enum class EventState : uint8_t
+    {
+      Invisible,
+      Available,
+      Locked
+    };
+
     using SceneReturnCallback = std::function< void( Entity::Player&, const SceneResult& ) >;
     using SceneChainCallback = std::function< void( Entity::Player& ) >;
     using EventFinishCallback = std::function< void( Entity::Player&, uint64_t ) >;
--- a/src/world/Event/EventHandler.h
+++ b/src/world/Event/EventHandler.h
@@ -85,7 +85,7 @@ namespace Sapphire::Event
       QuestBattleDirector = 0x8006,
     };
 
-    enum class EventState : uint8_t
+    enum class QuestAvailability : uint8_t
     {
       Invisible,
       Available,
--- a/src/world/Event/EventHandler.h
+++ b/src/world/Event/EventHandler.h
@@ -78,6 +78,8 @@ namespace Sapphire::Event
       FcTalk = 0x001F,
       Adventure = 0x0021,
       DailyQuestSupply = 0x0022,
+      TripleTriad = 0x0023,
+      PreHandler = 0x0036,
       ICDirector = 0x8003,
       PublicContentDirector = 0x8004,
       QuestBattleDirector = 0x8006,
--- a/src/world/Manager/MapMgr.cpp
+++ b/src/world/Manager/MapMgr.cpp
@@ -1,0 +1,889 @@
+#include <Common.h>
+#include <Service.h>
+
+#include <datReader/DatCategories/bg/lgb.h>
+#include <Exd/ExdDataGenerated.h>
+
+#include <Event/EventHandler.h>
+
+#include <Network/CommonActorControl.h>
+#include <Network/GameConnection.h>
+#include <Network/PacketDef/Zone/ServerZoneDef.h>
+#include <Network/PacketWrappers/ActorControlSelfPacket.h>
+
+#include <Logging/Logger.h>
+#include <Util/Util.cpp>
+
+#include <Actor/Player.h>
+#include <Inventory/Item.h>
+
+#include "MapMgr.h"
+#include "TerritoryMgr.h"
+
+#include "ServerMgr.h"
+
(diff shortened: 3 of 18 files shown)


src/world/Manager/MapMgr.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+  auto& quest = m_questData[ eventData.first ];
+
+  if( ( player.isQuestCompleted( eventData.first ) && ( !quest.isRepeatable && eventData.first != 67114 ) ) || player.hasQuest( eventData.first ) ||
+      ( quest.repeatIntervalType == 1 && quest.questRepeatFlag == 0 ) ) // Don't show daily beast tribe quests on the map yet.
+    return false;
+
+  if( quest.classJobUnlock )
+  {
+    if( quest.column18 == 3 )
```
NotAdam, line 334: this should be named instead of using the unknown value

src/world/Manager/MapMgr.cpp
```diff
@@ -0,0 +1,889 @@
...
+  if( Common::CURRENT_EXPANSION_ID < quest.expansion )
+    return false;
+
+  if( quest.mount )
+  {
+    if( !player.hasMount( quest.mount ) )
+      return false;
+  }
+
+  if( eventData.first == 65968 ) // Quest: A Legend for a Legend
```
NotAdam, line 360: things like this and basically everything after it should be deferred to a script, not inline

src/world/Manager/MapMgr.cpp
```diff
@@ -0,0 +1,889 @@
...
+      }
+    }
+  }
+  std::cout << "\n";
+}
+
+void Sapphire::World::Manager::MapMgr::updateAll( Entity::Player& player )
+{
+  auto& mapData = m_mapData[ player.getTerritoryTypeId() ];
+  std::multimap< uint32_t, EventData, less > newMapData;
```
NotAdam, line 234: is there a particular reason for using a multimap instead of just a list with a defined structure?

src/world/Manager/MapMgr.h
```diff
@@ -0,0 +1,132 @@
...
+    
+  private:
+    struct EventData
+    {
+      uint32_t iconId;
+      uint32_t levelId;
+      uint32_t actorId;
+    };
+
+    struct QuestData
```
NotAdam, line 40: this should be using the struct in exddatagenerated

NotAdam (changes requested):
meant to get to this last week but have been busy, sorry

src/world/Manager/MapMgr.cpp
```diff
@@ -0,0 +1,889 @@
...
+                eventData.iconId = exdData.get< Data::EventIconType >( 5 )->mapIconAvailable + 1;
+
+                m_mapData[ territoryTypeId ].insert( std::make_pair( npcData, eventData ) );
+                break;
+              }
+
+              case EventHandler::EventHandlerType::CustomTalk:
+              {
+                // Include only the beginner arena icon yet. There a few other ones, that aren't referenced in the game files (Some examples are: The Triple Triad Tournament npc which has multiple icons and the ocean fishing icon)
+                if( npcData == 721223 )
```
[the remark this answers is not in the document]
Taezen: Sorry, but I do not understand where the problem here is.
The custom talk icons have a very special handling from the server and not all of them are referenced in the exd data (The server decides which icon is displayed on the map. This applies to all supported events). So I have included this icon for now only.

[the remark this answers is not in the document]
NotAdam: i think i started writing something a few days ago and then didn't finish, i think what i intended to write here is that this shouldn't be hardcoded in some random file but i can't be sure lol

src/world/Manager/MapMgr.cpp
```diff
@@ -0,0 +1,889 @@
...
+                //m_mapData[ territoryTypeId ].insert( std::make_pair( eventData, 0 ) );
+                break;
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+  std::cout << "\n";
```
NotAdam, line 228: also use the logger to write out instead of iostream

src/common/Network/PacketDef/Zone/ServerZoneDef.h
```diff
@@ -928,12 +928,12 @@ namespace Sapphire::Network::Packets::Server
...
     uint32_t unknown8;
-    uint16_t festivalId;
-    uint16_t additionalFestivalId;
     uint32_t unknown9;
     uint32_t unknown10;
+    uint32_t festivalId;
     uint32_t unknown11;
-    uint32_t unknown12[4];
+    uint32_t unknown12[3];
+    uint32_t additionalFestivalId;
```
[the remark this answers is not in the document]
Taezen: This is an InitZone packet from the Gold Saucer event:
`10 00 90 00 00 00 00 00 00 00 00 00 17 46 56 00`
`02 59 00 00 AA D8 CF 00 02 00 00 00 00 80 09 41`
`00 00 80 3F` **73 00 00 00** `00 00 00 00 00 00 00 00`
`00 00 00 00` **73 00 00 00** `00 00 00 00 00 00 00 00`
`00 00 00 00 59 1D 0D C2 F9 D5 23 3F B7 FF C7 42`
`00 01 01 04 04 01 04 03 01 04 02 00 00 00 00 00`

This is an InitZone packet from the Moonfaire event:
`17 00 89 00 00 00 00 00 00 00 00 00 C0 1E 81 00`
`02 39 10 00 AA E0 24 00 15 00 00 00 00 80 09 41`
`00 00 80 3F` **74 00 00 00** `00 00 00 00 00 00 00 00`
`00 00 00 00 `**74 00 00 00** `00 00 00 00 00 00 00 00`
`00 00 00 00 7B 94 75 43 2B DA 97 42 93 85 4D 44`
`00 01 04 04 01 01 01 04 01 02 01 00 00 00 00 00`

Because of these two festival InitZone packets, I think they have changed something in the festival system.

src/world/Manager/MapMgr.cpp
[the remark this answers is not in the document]
Taezen: Aah, you're right, that this should'nt be hardcoded, but I think this will do it for now, until it is figured out which custom talk icons are shown by retail. In addition it is unclear for me now which icons are displayed for custom talk npcs that aren't referenced in the exd data.

src/world/Manager/MapMgr.cpp
```diff
@@ -0,0 +1,889 @@
...
+  auto& exdData = Common::Service< Data::ExdDataGenerated >::ref();
+
+  size_t count = 0;
+
+  for( auto quest : exdData.m_QuestDat.get_rows() )
+  {
+    if( exdData.getField< std::string >( quest.second, 1 ).empty() ) // id
+      continue;
+
+    auto& questData = m_questData[ quest.first ];
```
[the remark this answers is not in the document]
Taezen: I've tried it in this way first, but it takes approximately 10-20 ms to load the quest data directly from the exd. This is way too long, if you want to do it after the ZoneIn trigger like the retail server does.
Even if I remove the unused vectors from the Quest struct, it isn't really faster.
There are zones which contains over 100 quests to check. That's over a second processing time.
I didn't found any other way to this faster without storing them in memory.

src/world/Manager/MapMgr.cpp
```diff
@@ -0,0 +1,889 @@
...
+      m_mapData[ territory ].emplace( quest.first, eventData );
+    }
+  }
+  
+  auto& serverMgr = Common::Service< World::ServerMgr >::ref();
+  auto m_gameData = new xiv::dat::GameData( serverMgr.getConfig().global.general.dataPath );
+
+  std::set< std::string > bgSet;
+
+  for( auto territoryTypeId : exdData.getTerritoryTypeIdList() )
```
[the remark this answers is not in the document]
Taezen: I didn't see any event nor npc data in the InstanceObjectCache or did I miss something?

src/world/Manager/MapMgr.cpp
Taezen: Ok I change that, but I thought this is ok, because in the `InstanceObjectCache` initialization it's done in the same way.

src/world/Manager/MapMgr.cpp
Taezen: The retail server seems to sort it by the lower word of the event id.
So I used a multimap for auto sorting with a custom sort method.
For lower processing time we can use a list here, because it seems the sorting is irrelevant for the client.

src/world/Manager/MapMgr.cpp
Taezen: like an availability check in a quest script?

NotAdam: yeah, probably something like `bool showMapIcon(Entity::Player&)`

src/world/Manager/MapMgr.cpp
NotAdam: yeah my main concern was that this obscures the behavior and makes it a bit more complicated for no real gain - if ordering was necessary though i would just opt to store them with the event id as they'll be implicitly ordered by that anyway if they're dumped into a set/map when the key is the event id

src/world/Manager/MapMgr.cpp
NotAdam: more just some debugging code that was left behind lol

src/world/Manager/MapMgr.cpp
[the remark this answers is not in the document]
NotAdam: ah, they're filtered out - you can add them in here https://github.com/SapphireServer/Sapphire/blob/develop/src/world/Territory/InstanceObjectCache.cpp#L80

src/world/Manager/MapMgr.cpp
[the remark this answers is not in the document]
NotAdam: ah yeah, okay, i'm not surprised, datreader needs more than a bit of love in some areas

you can keep them in memory though, its just loading everything upfront is a bit costly when working on the server - i don't mind it that much as when i tried it, it wasn't too bad but its nice to avoid doing everything upfront where reasonable

Taezen:
I have changed it to actually use the struct from `ExdDataGenerated.h` for the quest data.
It now caches only the quest data during initialization. All other is generated during runtime or fetched from the `InstanceObjectCache`.
I also removed the hardcoded quest checks and implemented a basic method for an availability check.
Unfortunately I had to hardcode the data for `GuildLeveAssignment`, because exd data isn't accessible from scripts right now.

src/world/Event/EventHandler.h
```diff
@@ -85,6 +85,13 @@ namespace Sapphire::Event
       QuestBattleDirector = 0x8006,
     };
 
+    enum class EventState : uint8_t
```
NotAdam, line 88: i feel like `EventState` is a bit ambiguous, perhaps `EventAvailability`?

src/world/Script/NativeScriptApi.cpp
```diff
@@ -138,6 +137,11 @@ namespace Sapphire::ScriptAPI
   {
   }
 
+  Event::EventHandler::EventState EventScript::checkExtraConditions( Sapphire::Entity::Player& player, uint32_t eventId )
```
NotAdam, line 140: probably want this to replicate the name of the struct a bit more, `checkExtraConditions` doesn't really mean anything obvious

Taezen: I think this should be a function that checks requirements that aren't in the exd data. So this should called after the generic availability check.
So are something like `checkSpecialRequirements` are a better name for this?
Function names aren't my strength xD

NotAdam: i thought about this a bit and i would actually prefer that this is just called `getQuestAvailability` and the return is just like `QuestAvailability` - i can't think of any cases where this would apply to anything that isn't a quest in this case, and we can call different handlers for different cases as needed

i'd rather have more specific handlers that have a very specific scope rather than one general function that does a bunch of different things because that'll just be a nightmare over time

Taezen: Should this enum then stay in `EventHandler` or should I rather move it to `Common`?

NotAdam: nah its fine where it is, it doesn't really make sense for it to be in common anyway given its only really used within the event system

NotAdam (approved):
looks good - nice work

Taezen:
Thanks :)

PR #728, opened 2021-09-13
exd parser optimizations + support for sub rows in get_rows()

Taezen:


deps/datReader/Exd.cpp
```diff
@@ -279,106 +233,100 @@ namespace xiv::exd
   }
 
   // Get all rows
-  const std::map< uint32_t, std::vector< Field>>& Exd::get_rows()
+  const std::map< double, std::vector< Field > > Exd::get_rows()
```
NotAdam, line 236: yeah, uh, lets not do this lol, this is bound to have some bullshit issue because of floating point inaccuracy

just have a u64 and treat the low/high u32 as the row/subrow

NotAdam: or even use a struct and abuse a union to give you a u64 and 2 u32s

Taezen: The floating point inaccuracy wasn't really a problem in testing, but as you said it could create problems in the future.
Unfortunately unions aren't working as a key in containers. So I will went with your first approach.

NotAdam: you will need to add a custom hash function and then it'll work

NotAdam: https://en.cppreference.com/w/cpp/utility/hash

NotAdam: but you can just pass it the u64 value directly

Taezen: I could do that, but I don't wanted to make it more complex as necessary

NotAdam: it is generally helpful to be able to access the row ids easily, so it being simpler here makes it more complicated elsewhere ;)

Taezen: yeah, you're right 👍

NotAdam:
cool, thanks. tbh this needs to be rewritten as there's far better ways of doing this but this works better anyway

PR #729, opened 2021-09-13
Update opcodes for 5.58 hotfix and fix CFCommenceHandler

Skyliegirl33:
This should fix the Duty Finder and update the opcodes for the newest client.

The opcode updates are courtesy of collett, Flavien Normand and Taezen.
I had assistance with CFCommenceHandler from Mordred and collett.

### DIFF
--- a/src/common/Network/PacketDef/Ipcs.h
+++ b/src/common/Network/PacketDef/Ipcs.h
@@ -5,7 +5,7 @@
  
 namespace Sapphire::Network::Packets
 {
- 
+
   ////////////////////////////////////////////////////////////////////////////////
   /// Lobby Connection IPC Codes
   /**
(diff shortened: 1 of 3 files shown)


NotAdam:
can you fix the extra whitespace that was added, likely by an autoformatter?

also not sure if you care, but your git email is set incorrectly

Skyliegirl33:
That should have removed the whitepaces. Also, is it? My email is private, but I thought it would use the GH provided noreply email.

NotAdam:
it was using `From: Skyliegirl33 <{ID}+{username}@users.noreply.github.com>` ;)

Skyliegirl33:
Ahh, haha. I'll fix that

PR #730, opened 2021-09-14
fix the loading of some quest battles

collett8192:


src/world/Manager/TerritoryMgr.cpp
```diff
@@ -302,7 +302,26 @@ Sapphire::TerritoryPtr Sapphire::World::Manager::TerritoryMgr::createQuestBattle
...
-  auto pQuestInfo = exdData.get< Sapphire::Data::Quest >( pQuestBattleInfo->quest );
+  auto questId = pQuestBattleInfo->quest;
+
+  if( questId > 0x1FFFF )
+  {
+    auto arrayEventHandler = exdData.get< Sapphire::Data::ArrayEventHandler >( questId );
+    if( arrayEventHandler )
+    {
+      int i = 0;
+      while( i < arrayEventHandler->data.size() )
```
NotAdam, line 313: this is just a for loop with extra steps?

src/world/Manager/TerritoryMgr.cpp
```diff
@@ -302,7 +302,26 @@ Sapphire::TerritoryPtr Sapphire::World::Manager::TerritoryMgr::createQuestBattle
   if( !pQuestBattleInfo )
     return nullptr;
 
-  auto pQuestInfo = exdData.get< Sapphire::Data::Quest >( pQuestBattleInfo->quest );
+  auto questId = pQuestBattleInfo->quest;
+
+  if( questId > 0x1FFFF )
```
NotAdam, line 307: shouldn't this check `EventHandlerType` too?

PR #732, opened 2021-09-16
Various fixes, gear set unlocking and added ability to unlock mounts / orchestrion rolls

Skyliegirl33:
The `EquippedMannequin` db field is now used to determine how many gear sets are allowed to the character - this will have to be shared with the current gear set's index once I make a PR for that. Some actor controls were incorrect and missing, and two additional structs are now added. Other than that, there's just some new getters and setters for things I felt were missing, and some methods have an optional argument to not send update packets immediately since some functions shouldn't have that. `insertClassIntoDb` is also not hardcoded to level 1 anymore.


Skyliegirl33:
Oops. Orchestrion item doesn't work right, so ignore that for now. I deleted it

collett8192:
If you don't mind I'd like to take part of this PR and combine them into a later batch of 5.58 updates.

Skyliegirl33:
I don't mind if you keep me as the commit author or mention it somewhere at least

collett8192:
Sure thing, as in my every "stealed" commit.

collett8192:
Here we go https://github.com/SapphireServer/Sapphire/pull/876/commits/ea27ff3a55c29ecdeddbe41215c7ecde0207d531
Messed up the commits in the first try which left multiple mentions above.

Skyliegirl33:
Looks good!

PR #828, opened 2023-01-16
[3.x] fix nav_export for 3.05

takhlaq:
- re-add multithreaded exporting and make it eat less ram


T00fy:
FWIW this PR works for me fine on linux

PR #829, opened 2023-01-17
Fixing the starting Ul'dah quest (ManWil004) by using the new bindSceneReturn API 

T00fy:
Fixed in a similar way to the changes done for ManFst002


NotAdam (changes requested):
can you fix the syntax/formatting (spacing)

T00fy:
Updated with better formatting

PR #846, opened 2023-01-21
[3.3] Fix Emote exd struct

Skyliegirl33:


Skyliegirl33:
Added fixes for CustomTalk and Mount as well

PR #861, opened 2023-02-04
scripts: build modules to tmp directory before copying over to reload

takhlaq:
fixes infinite log spam


T00fy:
Verified that this PR works for me fine on linux

PR #878, opened 2023-02-07
Fixing EventItem struct for 3.3. Used quest ClsThm050 as a test

T00fy:


deps/datReader/Exd/Structs.h
```diff
@@ -491,13 +491,13 @@ namespace Excel
...
-    int32_t CastTimeline;
-    int32_t Timeline;
     uint16_t Icon;
     uint16_t Action;
     uint8_t StackMax;
+    uint8_t padding0[1];
     uint8_t CastTime;
-    int8_t padding1[2];
+    uint8_t Timeline;
+    int32_t CastTimeline;
```
SapphireMordred, line 500: Unfortunately this is not correct. i just had a brief look into the struct.
It appears as if they split it up now.  Before everything was in one struct, now it is split into EventItem, EventItemCastTimeline and EventItemTimeline.

PR #881, opened 2023-02-08
FreeCompanyMgr: fix invalid expression for *nix

Skyliegirl33:


SapphireMordred:
I am going with an entirely different solution here.

PR #883, opened 2023-02-09
Reimplemented EventItem. Fixed bug with onDeath not triggering when b…

T00fy:
…npc kills player. Fixed EventItem struct. Implemented ClsThm050 quest. Fixed some confusing and wrong param names

### DIFF
--- a/deps/datReader/Exd/Structs.h
+++ b/deps/datReader/Exd/Structs.h
@@ -493,8 +493,8 @@ namespace Excel
     uint32_t EventHandler;
     uint16_t Icon;
     uint16_t Action;
-    uint8_t StackMax;
     uint8_t Unknown;
+    uint8_t StackMax;
     uint8_t CastTime;
     uint8_t CastTimeLine;
     uint8_t Timeline;
--- a/deps/datReader/Exd/Structs.h
+++ b/deps/datReader/Exd/Structs.h
@@ -498,7 +498,7 @@ namespace Excel
     uint8_t CastTime;
     uint8_t CastTimeLine;
     uint8_t Timeline;
-    int8_t padding1[3];
+    uint8_t padding1[3];
   };
 
   struct EventItemCastTimeLine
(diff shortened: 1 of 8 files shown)


deps/datReader/Exd/Structs.h
```diff
@@ -494,9 +494,11 @@ namespace Excel
     uint16_t Icon;
     uint16_t Action;
     uint8_t StackMax;
+    int8_t padding0[1];
```
SapphireMordred, line 497: Are you sure about this? for one, in the exd this is an unsigned field, also you sure this isn't already the casttimeline?

T00fy: putting a byte here was the only way i could get the values to marry up nicely, otherwise CastTime never gets the correct value. 

But yeah, I could make it a uint.

T00fy: @SapphireMordred are you suggesting that it is CastTimeline that should be used instead of CastTime? In my quest if I didn't put the padding in then CastTime = 1 and CastTimeLine = 3. 
I checked a video online for the quest and the cast should take 3 seconds, that's why I am confused.

SapphireMordred: What i am saying is that it should just be uint8_t Unknown; instead of int8_t padding ;)

T00fy: haha ok, fair enough. Changed :)

deps/datReader/Exd/Structs.h
```diff
@@ -494,9 +494,11 @@ namespace Excel
     uint16_t Icon;
     uint16_t Action;
     uint8_t StackMax;
+    uint8_t Unknown;
```
SapphireMordred, line 497: Okay, almost! Now if you switch unknown and stack max, i'll be happy

T00fy: :+1:  done

PR #906, opened 2023-03-02
6.31 hotfix forks merge to official sapphire

Reyli:
Took able to walk around in Endwalker content from unofficial forks and want to be able to target retail with this branch.


src/world/Actor/PlayerQuest.cpp
```diff
@@ -1083,29 +1083,22 @@ bool Sapphire::Entity::Player::giveQuestRewards( uint32_t questId, uint32_t opti
...
+        addItem( itemId, questInfo->itemCountReward.at( i ), false, false, true, true );
       }
     }
   }
 
-  auto isHq = false;
-  if( optionalChoice > 1000000 )
-  {
-    optionalChoice -= 1000000;
-    isHq = true;
```
collett8192, line 1098: I think you want to keep this, unless it is actually changed in EW.

Reyli: I tried tracking down where that originated. And looked at the blame for it. You're definitely right about keeping it.

PR #937, opened 2023-05-23
Update Retail to 6.40

MapleHinata:
Client can login into the World, but ClientZone is wrong and so is the PlayerSetup struct, i'll fix it later, for now i'm leaving this as a draft.


Infiziert90:
MarketBoardPurchaseHandler isn't resolving correctly for me. 
My Result is 0x256

MapleHinata:
I think this PR is ready to be merged, everything that used to work on 6.3x, should work now.
2023-05-27_00-12-32-534_Thissucks

PR #946, opened 2023-07-18
5.58 action update

collett8192:
merge main 5.58,
update WAR to 5.58,
fix player combat state.

### DIFF
--- a/src/world/Actor/Chara.cpp
+++ b/src/world/Actor/Chara.cpp
@@ -355,16 +355,13 @@ bool Sapphire::Entity::Chara::checkAction()
 
 void Sapphire::Entity::Chara::update( uint64_t tickCount )
 {
-  if( isAlive() )
+  updateStatusEffects();
+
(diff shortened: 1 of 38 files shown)


collett8192:
Mob movement fix added.

collett8192:
WHM updated

src/world/Actor/Chara.cpp
```diff
@@ -355,13 +355,16 @@ bool Sapphire::Entity::Chara::checkAction()
 
 void Sapphire::Entity::Chara::update( uint64_t tickCount )
 {
-  updateStatusEffects();
-
-  if( std::difftime( static_cast< time_t >( tickCount ), m_lastTickTime ) > 3000 )
+  if( isAlive() )
```
hkAlice, line 358: this will work for intended purposes, but some status effects still 'tick' even if the actor is dead (eg. ifrit searing wind)

collett8192: Yeah, I updated it to only stop dot/hot.

PR #950, opened 2023-10-20
Login.html refactored

Sveimoldr:
The JavaScript code has been moved to separate files: "assets/js/login.js" and "assets/js/github.js". Also, event listeners have been added to handle the login functionality and fetch recent activity from the GitHub API.


NotAdam:
but... why?

PR #956, opened 2024-06-10
Update to 6.58 hotfix 2

collett8192:
PlayerSetup needs more love, ~~anything below "name" is not updated.~~
~~Missing some quest and event opcodes, new characters won't work since they are used in the opening event.~~

Opcode mainly stealed from https://github.com/karashiiro/FFXIVOpcodes

Edited: more fields in PlayerSetup updated.
Edited2: quests working now


collett8192:
The non-action 5.58 branch is merged into this PR now thus can be deleted.

PR #958, opened 2024-06-20
[3.x] WIP: Initial Action and StatusEffect implementation

arieshi255:
Still a WIP, but can be useful for some other things, so submitting this now.
Development will be done on `want_some_action`.


src/world/Action/Action.cpp
```diff
@@ -583,13 +590,72 @@ void Action::Action::buildActionResults()
...
+  }
+}
+
+void Action::Action::handleJobAction()
+{
+  switch( m_pSource->getClass() )
+  {
+    case ClassJob::Warrior:
+    {
+      Warrior::onAction( *m_pSource->getAsPlayer(), *this );
```
NotAdam, line 653: i'm okay with this as is for now, but i think it would probably better to have an array or otherwise that's just indexed by classjob (which also lets you pick up the base classjob impl easy) instead of this - i get this is wip-ish though so i'm not too worried about it at the moment

deps/datReader/Exd/Structs.h
```diff
@@ -430,7 +430,7 @@ namespace Excel
     uint8_t HideCastBar : 1;
     uint8_t IsTargetLine : 1;
 
-    int8_t padding0;
+    int8_t unknown : 8;
```
SapphireMordred, line 433: why : 8 ? it is already an 8 bit value.

src/scripts/statuseffect/StatusEffectDefiance.cpp
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  void onExpire( Entity::Chara& actor ) override
+  {
+    actor.removeSingleStatusEffectById( Unchained );
+    actor.removeSingleStatusEffectById( Wrath );
+    actor.removeSingleStatusEffectById( WrathII );
+    actor.removeSingleStatusEffectById( WrathIII );
+    actor.removeSingleStatusEffectById( WrathIV );
+    actor.removeSingleStatusEffectById( Infuriated );
```
SapphireMordred, line 23: it might be better to have something that can remove a list of ids if that is required.

src/world/Action/Action.cpp
```diff
@@ -583,13 +590,72 @@ void Action::Action::buildActionResults()
       shouldRestoreMP = false;
     }
   }
+
+  // If we hit an enemy
+  if( m_hitActors.size() > 0 && getHitChara()->getObjKind() != m_pSource->getObjKind() )
```
SapphireMordred, line 595: i'd prefer !m_hitActors.empty(), it is more elaborate

src/world/Action/Action.cpp
```diff
@@ -583,13 +590,72 @@ void Action::Action::buildActionResults()
...
+  {
+    m_pSource->removeStatusEffectByFlag( Common::StatusEffectFlag::RemoveOnSuccessfulHit );
+  }
+
+  handleJobAction();
+
+  if( m_lutEntry.statuses.caster.size() > 0 || m_lutEntry.statuses.target.size() > 0 )
```
SapphireMordred, line 602: same thing here. !empty() is cleaner than size() > 0

src/world/Action/ActionResult.cpp
```diff
@@ -89,6 +101,19 @@ void ActionResult::applyStatusEffectSelf( uint32_t id, int32_t duration, uint8_t
   m_pStatus->setParam( param );
 }
 
+void ActionResult::applyStatusEffectSelf( uint32_t id, int32_t duration, uint8_t param, std::vector< World::Action::StatusModifier > modifiers,
```
SapphireMordred, line 104: as mentioned, i think it would be good to not pass a copy of the vector here.

src/world/Actor/Chara.cpp
```diff
@@ -536,6 +535,18 @@ void Chara::statusEffectFreeSlot( uint8_t slotId )
...
+{
+  for( const auto& effectIt : m_statusEffectMap )
+  {
+    if( effectIt.second->getId() == id )
+    {
+      removeStatusEffect( effectIt.first, false );
+      break;
+    }
+  }
+}
```
SapphireMordred, line 548: This is a map, why do you iterate it to find an id? You should be able to jund "find()" it.

src/world/StatusEffect/StatusEffect.cpp
```diff
@@ -125,6 +191,21 @@ uint64_t Sapphire::StatusEffect::StatusEffect::getStartTimeMs() const
   return m_startTime;
 }
 
+uint32_t Sapphire::StatusEffect::StatusEffect::getFlag() const
+{
+  return m_flag;
+}
+
+std::vector< World::Action::StatusModifier > Sapphire::StatusEffect::StatusEffect::getStatusModifiers() const
```
SapphireMordred, line 199: this should probably return a const ref

PR #972, opened 2025-01-17
[3.x] wip: Fix Duty Finder to work again

arieshi255:


src/world/ContentFinder/ContentFinder.cpp
```diff
@@ -160,9 +161,13 @@ void World::ContentFinder::completeRegistration( const Entity::Player &player, u
   if( flags & 0x01 )
   {
     auto updatePacket = makeUpdateFindContent( player.getId(), content->data().TerritoryType,
-                                               CompleteRegistration, 0x20, static_cast< uint32_t >( player.getClass() ) );
+                                               CompleteRegistration, 1, static_cast< uint32_t >( player.getClass() ), 0x20 );
```
SapphireMordred, line 164: better not use magic numbers here but a variable name so we know what it means, otherwise this knowledge will probably get lost.

PR #974, opened 2025-05-07
Update action target filtering and add more calls to action script at different stages during execution.

collett8192:


Kooper16:
bumping this so it maybe gets checked on and conflicts get resolved. The new script calls would potentially allow me to implement Barrage and fix Wanderer's Minuet

collett8192:
conflict solved

PR #982, opened 2025-07-23
Implemented BRD skills for level <= 15 & some associated server features

Kooper16:
The associated server features are:

1. Crit modifiers from the user are now factored into the crit rate (for Straight Shot)
2. onSkillProc function inside PlayerMgr.cpp added for skill procs
3. onPlayerHit status effect script executed inside the damage function of ActionResultBuilder so we can drop the Bind status effect on hit.
4. Fixed auto attack skill for BRD and MCH

This PR also includes https://github.com/SapphireServer/Sapphire/pull/978 and https://github.com/SapphireServer/Sapphire/pull/979


Kooper16:
I'm really sorry for all these changes. I'll make smaller PRs next time QQ

src/scripts/statuseffect/detrimental/StatusEffectBind.cpp
```diff
@@ -0,0 +1,25 @@
+#include <Script/NativeScriptApi.h>
+#include <ScriptObject.h>
+#include "StatusEffect/StatusEffect.h"
+#include <action/Action.h>
```
NotAdam, line 4:
```suggestion
#include <Action/Action.h>
```

to fix linux build

src/world/Script/NativeScriptApi.cpp
```diff
@@ -62,7 +62,7 @@ namespace Sapphire::ScriptAPI
   {
   }
 
-  void StatusEffectScript::onPlayerDamaged( Entity::Chara& actor )
+  void StatusEffectScript::onPlayerHit( Entity::CharaPtr& attackTarget, Sapphire::StatusEffect::StatusEffect& effect )
```
NotAdam, line 65: i assume the intention here would've been entity::chara& instead of entity::charaptr&? passing a reference to the pointer is a bit moot (and scripts shouldn't ever get ownership either)

PR #983, opened 2025-07-25
[3.x] CastType enum fix, collision filter cleanup, box filter and target filters in LUT

arieshi255:


NotAdam:
not sure we need two copies of the lut haha, probs delete the one in the src/ folder

PR #984, opened 2025-07-25
Quick nock & Cone aoe implementation

Kooper16:
- Implemented a system that reads shape parameters from a json for easy editing. As far as I know, there is not enough information inside the exd files to export all necessary parameters.
- Changed actor filters to require all of them to be true instead of just one for an actor to get hit by an action.
- The cone calculation need 2 parameters: The start angle (how much to the right of the player does the cone start) and the end angle (how much to the right of the player does the cone end). A 90° frontal cleave would be -45;45 or 315;45.
- The cone calculation has 2 parts: 
  1. First we do a radius check to see if the target is even close enough to be hit. 
  2. Then we calculate the angle from source to target relative to the X-axis and then shift all values by the same angle but towards the target of the skill. This allows our cone check to be in world rotation (which our start and end angles for the cone are). Now we simply check if the angle towards the target is between the start and end angles of the source.

- The range of quick nock is currently only 8y because the trait for the increased range is not implemented
- This issue contains https://github.com/SapphireServer/Sapphire/pull/982

I'm going to leave this as a draft for now because people on discord seem to consider changing how we do collisions in general but I wanted to leave this up for review anyway in case we still want something useable in the meantime. I will change this to a normal PR once the other PR got accepted.


Kooper16:
After thinking about this for a bit, I think it would be worth adding this PR either way because the biggest chunk of new/changed code is stuff that we are going to reuse if we decide on using a separate collision detection engine. Most of the changes from this PR are the new classes that allow us to store shape parameters in a json so we can then read them out for aoe shape calculations later. We are still going to need this data if we change the engine. The specific part that would get replaced (the actual cone aoe calculation) is relatively small but useful in the meantime so we can keep implementing cone aoe skills.

src/world/Util/ActorFilter.cpp
```diff
@@ -44,3 +45,31 @@ bool Sapphire::World::Util::ActorFilterBox::conditionApplies( const Sapphire::En
...
+  m_startPos( startPos ), m_skillTargetPos( skillTargetPos ), m_startAngle( startAngle ), m_endAngle( endAngle )
+{
+}
+
+bool Sapphire::World::Util::ActorFilterCone::conditionApplies( const Entity::GameObject& actor )
+{
+  Common::FFXIVARR_POSITION3 targetPos = actor.getPos();
+  
+  float angleToCurrentTarget = Sapphire::Common::Util::radianToDegrees( Sapphire::Common::Util::calcAngTo( m_startPos.x, m_startPos.z, targetPos.x, targetPos.z ) );
+  float angleToSkillTarget = Sapphire::Common::Util::radianToDegrees( Sapphire::Common::Util::calcAngTo( m_startPos.x, m_startPos.z, m_skillTargetPos.x, m_skillTargetPos.z ) );
```
Pythyu, line 60: I'm curious about why we need to convert it to degree ? Can't we set `m_endAngle` as radian when initialized so we don't have to do the conversion on every condition check ?

Kooper16: Good point. I will do that

src/scripts/statuseffect/brd/StatusEffectSwiftsongAura.cpp
```diff
@@ -0,0 +1,91 @@
...
+    {
+      actor.removeSingleStatusEffectById( SwiftsongAura );
+      return;
+    }
+
+    auto pPlayer = actor.getAsPlayer();
+    if( !pPlayer )
+      return;
+
+    if (pPlayer->isInCombat())
```
takhlaq, line 42: style
```cpp
if( pPlayer->isInCombat() )
```

src/scripts/statuseffect/brd/StatusEffectSwiftsongAura.cpp
```diff
@@ -0,0 +1,91 @@
...
+      return;
+    }
+
+    if( pPlayer->getPartyId() == 0 )
+      return;
+
+    auto& partyMgr = Common::Service< World::Manager::PartyMgr >::ref();
+    auto pParty = partyMgr.getParty( pPlayer->getPartyId() );
+    auto membersIds = pParty->MemberId;
+    for( auto& target : actor.getInRangeActors(false) )
```
takhlaq, line 54: style
```cpp
for( auto& target : actor.getInRangeActors( false ) )
```

src/world/WorldServer.cpp
```diff
@@ -263,6 +263,11 @@ void WorldServer::run( int32_t argc, char* argv[] )
     Logger::fatal( "Unable to cache actions!" );
     return;
   }
+  if (!pActionMgr->cacheActionShapeLut())
```
takhlaq, line 266: style
```cpp
if( !pActionMgr->cacheActionShapeLut() )
```

src/world/Actor/Player.cpp
```diff
@@ -603,7 +603,6 @@ bool Player::isInCombat() const
 
 void Player::setInCombat( bool mode )
 {
-  //m_lastAttack = GetTickCount();
   m_bInCombat = mode;
```
NotAdam, line 606: should this instead just check PlayerCondition instead? seems a bit redundant to have a bool and the condition

Kooper16: I just used what we have for this. I can change that if you want

PR #997, opened 2025-08-01
[3.x] Timeline implementation, Ifrit Trial

hkAlice:
### Summary
This PR implements a concept for scripting battle behavior in trials, raids, dungeon bosses, etc. It helps define a "battle" timeline in JSON, alongside an UI editor. We implement Ifrit as example.
It also implements concepts such as subactors (helper actors for complex mechanics), and selectors (selecting and filtering actors based on criteria related to aggro, positioning, ally, player role, status effect, etc).

Included are also many many fixes related to instance flow, resetting, director behavior, BattleTalk (npc dialogue) as well as BNPC control.

### Definition

EncounterTimelines are defined in data/EncounterTimelines. Our goal was to get Ifrit functional enough.

https://github.com/user-attachments/assets/a65272f4-8571-4fea-90fd-556fecf18f11

The timeline is currently defined by an external UI editor developed outside this repo: https://github.com/hkAlice/sapphire_editor

To edit the Ifrit timeline, either edit the JSON directly or use the tooling provided in https://hkalice.github.io/sapphire_editor/

image

- Actors are the BNPCs we intend on controlling, keyed by layoutID;
- Conditions are ways to "push" and control the battle or BNPC behavior, akin to a state machine (based on FFXII gambits);
- Schedules are grouped battle behavior on a "timeline", where timepoints are defined at a given t;
- Timepoints are an entry for a timeline to execute something (make a BNPC cast, change director vars, position a mob, etc);
- Selectors are a way to define a query to select targets, in which you can then use them on timepoints.

### Future goals

This PR has intention of providing more support and explanation when the feature has further matured to support a wider array of timepoints and manipulation necessary, and step two of it would be implementing another fight.

BNPC ownership is an issue, where the timeline should only "play" with the director given and BNPCs spawned. We need another layer or interface for indicating BNPC spawns, controlling the way they spawn (with VFXs), etc.


hkAlice:
Thanks to @takhlaq this passes CI and no merge conflicts

Feedback/review welcome

PR #999, opened 2025-08-04
CMake modernization and address sanitizer support

mrexodia:
- Bumped minimum CMake to 3.22 (Ubuntu 22.04 LTS)
- Add `ENABLE_SANITIZERS` option (needs wiki guide)
- Remove/simplify a bunch of compiler options (will add comments)
- Remove `project` calls (they should only exist if folders can be configured/built standalone)
- Everything builds (with and without sanitizers) on Visual Studio, bunch of errors at runtime fixed in #1000


cmake/compiler.cmake
```diff
@@ -1,37 +1,53 @@
...
-  add_definitions( -D_WIN32_WINNT=0x601 )
-  add_definitions( -D_CRT_SECURE_NO_WARNINGS )
-  add_definitions( -DNOMINMAX )
-      
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHc" )
-  message( STATUS "Enabling Build with Multiple Processes.." )
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4834" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /std:c++17" )
```
mrexodia, line 15: Removed because this is set with `CMAKE_CXX_STANDARD`

cmake/compiler.cmake
```diff
@@ -1,37 +1,53 @@
...
 
 if( UNIX )
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -fPIC" )
-  set( CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS} -O3")
-else()
-  add_definitions( -D_WIN32_WINNT=0x601 )
-  add_definitions( -D_CRT_SECURE_NO_WARNINGS )
-  add_definitions( -DNOMINMAX )
-      
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHc" )
```
mrexodia, line 10: This flag is generally weird, the default should be used unless there is a compelling reason.

PR #1000, opened 2025-08-04
Sanitizer fixes

mrexodia:
- [x] Requires #999 to be merged
- [x] Sanitizers are not linking correctly with clang
- [x] Use wildcard searching for MySQL libraries


mrexodia:
Most of these are unlikely to crash in release mode, but all of them are 100% bugs. The most critical one in the heap overflow in GamePacketParser, which would not detect incomplete packets and (likely) cause packets to be handled with random heap data.

PR #1008, opened 2025-08-06
Aggro implementation

Kooper16:
This PR will include a working aggro implementation that...

- [X] correctly calculates aggro including modifiers
- [x] streamlines aggro application for actions
- [x] streamlines aggro application inside status effects (eg. dot ticks)
- [X] applies aggro to all actors in combat with the target when using supportive actions

This PR also includes some bug fixes regarding aggro swap and displayed letters when entering combat.

Edit: PR is now ready. There will still be visual aggro bugs when a player enters from off screen and didn't get any aggro related packets yet (there are many issues outside of aggro regarding that). But if one would do a trial or raid together, then aggro management and display should be almost identical to retail.

### DIFF
--- a/src/common/Common.h
+++ b/src/common/Common.h
@@ -811,6 +811,14 @@ namespace Sapphire::Common
 
   };
 
+  struct QuestEobject
+  {
+    uint32_t id;
+    uint32_t teri;
+    FFXIVARR_POSITION3 pos;
+    float radius;
+  };
+
   enum class ParamModifier : uint16_t
   {
     None = 0,
--- a/src/scripts/quest/subquest/blackshroud_south/GaiUsa209.cpp
+++ b/src/scripts/quest/subquest/blackshroud_south/GaiUsa209.cpp
@@ -1,0 +1,255 @@
+// This is an automatically generated C++ script template
+// Content needs to be added by hand to make it function
+// In order for this script to be loaded, move it to the correct folder in <root>/scripts/
+
+#include <Actor/Player.h>
+#include "Manager/EventMgr.h"
+#include "Manager/PlayerMgr.h"
+#include <ScriptObject.h>
+#include <Service.h>
+
+// Quest Script: GaiUsa209_00732
+// Quest Name: Stash Saboteur
+// Quest ID: 66268
+// Start NPC: 1000598 (Auphiliot)
+// End NPC: 1000598 (Auphiliot)
+
+using namespace Sapphire;
+
+class GaiUsa209 : public Sapphire::ScriptAPI::QuestScript
+{
+  private:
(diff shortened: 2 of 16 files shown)


Kooper16:
> Fix aggro arrow/sfx not appearing when reaggroing a mob. It works if it switches target to another player or if the player reaggroing runs far enough to despawn to the mob and then runs back for the reaggro

Actually a feature and not a bug. YOSHIDAAAAAAAA

Kooper16:
> Fix enemies not appearing as untagged (but in combat). They currently get a purple name plate even though they are untagged

I currently believe next highest in aggro just gets red (tagged) instead. I will leave it like this and change it if needed

src/world/Actor/Chara.cpp
```diff
@@ -788,6 +788,12 @@ void Chara::setAgentId( uint32_t agentId )
   m_agentId = agentId;
 }
 
+std::vector< CharaPtr > Chara::getHateList()
+{
+  std::vector< CharaPtr > hateList = {};
+  return hateList;
```
hkAlice, line 794: prefer list initialization:

```cpp
return {};
```

src/world/Actor/Player.cpp
```diff
@@ -1025,6 +1052,18 @@ void Player::onMobDeaggro( const BNpc& bnpc )
   }
 }
 
+void Player::hateListLetterUpdate( const BNpc& bnpc )
+{
+  uint8_t hateId = m_actorIdTohateSlotMap[ bnpc.getId() ];
```
hkAlice, line 1057: prefer getting map by iterator instead of operator[] in this read instance

src/world/Actor/BNpc.cpp
```diff
@@ -615,29 +648,37 @@ void BNpc::aggro( const Sapphire::Entity::CharaPtr& pChara )
...
 {
   if( hateListHasActor( pChara ) )
+  {
     hateListRemove( pChara );
+    if( getTargetId() == pChara->getId() )
+    {
+      updateAggroTarget();
+      hateListUpdatePlayers();
+    }
+    if( m_pOwner == pChara  )
```
hkAlice, line 663: there's an extra space here

src/world/Action/ActionResultBuilder.h
```diff
@@ -47,9 +47,12 @@ namespace Sapphire::World::Action
 
   private:
     uint32_t m_actionId;
+    float m_aggroModifier;
     uint16_t m_requestId;
     uint32_t m_resultId;
 
+    bool m_applyHealAggro = true;
```
hkAlice, line 54: prefer initializing with {}

PR #1018, opened 2025-08-12
add ground target with flaming arrow

takhlaq:


takhlaq:
marked as draft until we come up with a generic solution for creating ground targets rather than scripting each case

PR #1023, opened 2025-08-15
add status effect display to party list

takhlaq:


hkAlice:
this seems like a clean and very small PR, was it tagged as draft by accident? or untested?

takhlaq:
Think I only tested this with characters in range (which can be updated via effect packet and actor controls). Can't remember if this actually worked or not for out of range members.

Will get back to this after encounters.

PR #1032, opened 2025-09-01
Add missing cstdint includes for the MySql/Database headers

redstrate:
This fixes builds on recent enough GCC versions where they culled whatever <cstdint> include was hiding in an already-included header.

### DIFF
--- a/deps/mysqlConnector/Connection.h
+++ b/deps/mysqlConnector/Connection.h
@@ -4,6 +4,7 @@
 #include <string>
 #include <memory>
 #include <map>
+#include <cstdint>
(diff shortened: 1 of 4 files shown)


deps/mysqlConnector/Connection.h
```diff
@@ -4,6 +4,7 @@
 #include <string>
 #include <memory>
 #include <map>
+#include <cstdint>
 #include "MysqlCommon.h"
```
mrexodia, line 8: wouldn't it make more sense to put it in "MysqlCommon.h"?

PR #1059, opened 2026-04-20
Timeline refactor and dummy Sastasha encounters

takhlaq:
Refactor timeline to use the new trigger and phases format.
Updated Sastasha with dummy encounters so the duty can be completed, fights still need scripting.


src/world/Encounter/TriggerCondition.cpp
hkAlice: small oops here:

```diff
-    m_flags = json.at( "flags" ).get< uint32_t >();
+    m_flags = paramData.at( "flags" ).get< uint32_t >();
```

src/world/Encounter/Trigger.cpp
```diff
@@ -0,0 +1,77 @@
...
+    return m_pCondition->isConditionMet( state, pack, pEncounter, time );
+  }
+
+  void Trigger::execute( TimelineActor& self, TimelinePack& pack, PhaseState& state, EncounterPtr pEncounter, uint64_t time ) const
+  {
+    switch( m_pAction->m_type )
+    {
+      case TriggerAction::Type::TransitionPhase:
+      {
+        self.transitionPhase( pEncounter, pack, time, self.getPhaseById( m_pAction->m_phaseId ) );
```
hkAlice: might be worth null checking here

```diff
-        self.transitionPhase( pEncounter, pack, time, self.getPhaseById( m_pAction->m_phaseId ) );
+       auto nextPhase = self.getPhaseById( m_pAction->m_phaseId );
+       if( !nextPhase )
+         // log, break, throw exception
+       else
+         self.transitionPhase( pEncounter, pack, time, nextPhase );
```

src/scripts/instances/trials/TheBowlofEmbers.cpp
```diff
@@ -61,7 +61,7 @@ class TheBowlofEmbers : public Sapphire::ScriptAPI::InstanceContentScript
       pEncounter->update( tickCount );
 
       // Fight end condition
-      if( pEncounter->getStatus() == EncounterStatus::ACTIVE )
+      if( pEncounter->getStatus() != EncounterStatus::SUCCESS )
```
hkAlice, line 64: is it worth checking for if it's active **and** success?

hkAlice (commented):
tiny bugfixes that are a bit invisible, mergeable as is

src/scripts/instances/trials/TheBowlofEmbers.cpp
takhlaq: left this as-is since if you one shot ifrit with gm command the encounter will lock (no combat state transition) as the encounter state would still be idle

PR #1060, opened 2026-04-21
Add Lua hybrid script runtime foundation

Aeshur:
Introduces the core plumbing needed to support Lua-backed scripts alongside the existing native script path

### DIFF
--- a/config/world.ini.default
+++ b/config/world.ini.default
@@ -1,10 +1,16 @@
 [Scripts]
+; which script runtime to prefer: native, lua, or hybrid
+Backend = native
 ; where compiled script modules are located
 Path = ./compiledscripts/
 ; relative to Path, where we copy and load modules from
 CachePath = ./cache/
 ; whether we should detect changes to script modules and reload them
 HotSwap = true
+; where experimental Lua content scripts live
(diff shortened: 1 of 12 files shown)


NotAdam:
firstly, it would've been nice if there was some discussion or otherwise before opening a PR to change/add a whole subsystem, but beyond that:
* not sure what the goal is of trying to parse lua scripts to read tables, this doesn't really handle syntax properly either at first glance and is brittle with no real error handling/recovery
* considerations with the existing native script system revolve around minimal dev/api surface friction, and changes are re-factorable across the whole codebase. adding lua throws a lot of that in the bin (which is _fine_), but there are better alternatives that provide arguably a nicer middleground (luau for example has type safety/hinting at the very least)
* lua scripts in this design silently get precedence over native scripts and there's no warning or otherwise which will just lead to confusion

changing how scripts work is absolutely on the table but this aint the way to go about it

Aeshur:
Fair enough, I converted to draft for now. I amended my previous changes to make the current behavior more predictable and switched hybrid mode so native remains authoritative during early adoption while Lua/native collisions are reported at startup and partial Lua module load failures now fail startup.

